PR #3, opened 2020-10-14
Enable GitHub Actions for this repository

johnmcollier:
Fixes https://github.com/devfile/api/issues/178

Enables GitHub Actions for CI in this repository, making sure that the index generator and registry docker image are build on each PR and push to master.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -11,10 +11,13 @@ jobs:
     name: Check index-generator build
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
+    - name: Check out code
+      uses: actions/checkout@v2
 
     - name: Setup Go environment
       uses: actions/setup-go@v2.1.2
+      with:
+        go-version: 1.13
 
     - name: Check if index-generator build is working
       run: cd index/generator && ./build.sh
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,30 @@
+name: CI
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  go:
+    name: Check index-generator build
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setup Go environment
+      uses: actions/setup-go@v2.1.2
+
+    - name: Check if index-generator build is working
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,30 @@
...
+jobs:
+  go:
+    name: Check index-generator build
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Setup Go environment
+      uses: actions/setup-go@v2.1.2
+
```
GeekArthur, line 18: It would be better we specify the go version here (like below) as we use `go mod` and the lower version of go might not work properly

```yaml
with:
        # The Go version to download (if necessary) and use. Supports semver spec and ranges.
        go-version: 1.13
```

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,30 @@
...
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+  go:
+    name: Check index-generator build
+    runs-on: ubuntu-latest
+    steps:
```
GeekArthur, line 13: nit: add `- name: Check out code` for this step is more clear and neat

index/generator/build.sh
```diff
@@ -1,7 +1,2 @@
 #!/bin/sh
-MODULE="github.com/devfile/registry-support/index/generator"
-BIN_DIR=$GOPATH/bin
-MODULE_BIN=$BIN_DIR/generator
-
-CGO_ENABLED=0 go install $MODULE
-cp $MODULE_BIN ./index-generator
+CGO_ENABLED=0 go build -mod=vendor -o index-generator .
```
GeekArthur, line 2: 👍

GeekArthur (changes requested):

.github/workflows/ci.yaml
johnmcollier: Sounds good 👍

.github/workflows/ci.yaml
johnmcollier: Ooh, good catch.

johnmcollier: Updated

.github/workflows/ci.yaml
johnmcollier: Updated

GeekArthur (approved):
LGTM, thanks!

PR #11, opened 2020-11-17
Use an nginx proxy server in the metadata server

johnmcollier:

Sets up a simple nginx reverse proxy on the metadata-server containe, allowing clients to connect to the OCI registry's v2 api  (running on port 5000 in the same pod) to pull devfiles. 

This PR also properly reformats the metadata-server container image to be used as a base image for the devfile stacks + index (see https://github.com/devfile/registry/pull/1, and https://github.com/devfile/api/blob/master/docs/proposals/registry/devfile-packaging.md). As a result, this container no longer packages the index.json or devfile stacks and is no longer meant to be deployed directly, but rather be used as a base image for the devfile index image.

https://github.com/devfile/api/issues/190, only partially (see last bullet point)


oci-devfile-registry-metadata/nginx.conf
```diff
@@ -0,0 +1,62 @@
...
+        # Temporary until the new metadata server has been delivered
+        # ToDo: Update to point to the new server when ready
+        default_type application/json;
+        root /www/data;
+        index index.json;
+    }
+
+    location /v2 {      
+      # Only allow HEAD (implicit) and GET requests on the V2 api
+      limit_except GET {
```
GeekArthur, line 45: Please correct me if I am wrong. Seems like the current nginx server works as reverse proxy and redirects:
- registry API requests to registry container (block all requests except GET requests)
- other requests to index server

Given our current index server already exposes the REST APIs (/index, /devfiles/{name}) to consumers to let consumers to GET index and devfile, in other words, it already implicitly blocks the PUT/POST/DELETE requests as we don't support them for now, for the time being, seems like the current index server has the same functionalities as using index server with nginx.

Also I am thinking more about the future work for supporting dynamic registry (support POST/PUT/DELETE) as I believe currently both index server and index server with nginx don't support those functionalities.
- If we use index server we can expose more REST APIs and design it to make it more user-friendly for devfile user, inside index server we still use ORAS package to interact with registry container (it's truly not accessible from public)
- If we use index server with nginx, we rely on registry API to let use interact with registry container directly within index server container, for this case I am not sure if it's easier to use as the GET registry API is `GET /v2/<name>/blobs/<digest>`, seems like user still need to pass/store some other information instead of `GET /devfiles/{name}`

johnmcollier: You're right, it is a reverse proxy, and the doc here: https://docs.docker.com/registry/recipes/nginx/ confirms that. I'll update the PR description.

The reason we still need to expose the OCI registry to clients is because more than just the devfiles will be stored in the OCI registry (e.g. plugins, stack artifacts, etc). Using the index server's `GET /devfiles/{name}` is good for retrieving the devfile, but it doesn't work well if other artifacts are stored as part of the stack (like binary blobs).

GeekArthur (approved):
LGTM, thanks for making the changes!

GeekArthur (approved):
Add LGTM back

PR #12, opened 2020-11-23
Implement index server

GeekArthur:
Signed-off-by: jingfu wang <jingfu.j.wang@ibm.com>

This PR implements the new index server, it also integrates with nginx server to serve non-registry API requests.

Fixes https://github.com/devfile/api/issues/188

**PR acceptance criteria**:

- [x] Test (WIP) 

- [ ] Documentation (WIP)

Deploy index server and registry as before, send GET request to endpoint /index or /devfiles/{devfile name}


index-server/push.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/sh
 IMAGE_TAG=$1
 docker tag devfile-index-base:latest $IMAGE_TAG
-docker push $1
\ No newline at end of file
+docker push $IMAGE_TAG
```
johnmcollier, line 4: Haha, good catch on your part 👍

index-server/Dockerfile
```diff
@@ -0,0 +1,42 @@
...
+    chown -R www-data:root /var/cache/nginx && \
+    chown www-data:root /etc/nginx/conf.d /etc/nginx/nginx.conf
+RUN chmod g+rwx /var/run/nginx.pid && \
+    chmod -R g+rwx /var/cache/nginx && chmod -R g+rwx /etc/nginx && chmod -R g+rwx /www/data
+
+USER www-data
+
+# Set env vars for the locations of the devfile stacks and index.json
+ENV DEVFILE_STACKS /registry/stacks
+ENV DEVFILE_INDEX /registry/index.json
```
johnmcollier, line 39: Same as my other comment, can you update `DEVFILE_STACKS` and `DEVFILE_INDEX` to use `/stacks` and `/index.json` as that's what https://github.com/devfile/registry is using?

index-server/entrypoint.sh
```diff
@@ -0,0 +1,22 @@
...
+if [ ! -e "$DEVFILE_INDEX" ]; then
+    echo "The container does not contain an index.json at $DEVFILE_INDEX. Exiting..."
+    exit 1
+fi
+
+# Start the nginx server
+nginx -g "daemon off;" &
+
+# Start the index server
+/registry/index-server
```
johnmcollier, line 20: Curious, why do we start the nginx proxy first, rather than the index server? My thinking was something like this:
```
# Start the index server
/registry/index-server &

# Start the nginx server
exec "$@"
```

(Both should work fine, just curious)

index-server/main.go
```diff
@@ -0,0 +1,151 @@
...
+	"github.com/gin-gonic/gin"
+	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
+)
+
+const (
+	devfileName            = "devfile.yaml"
+	devfileConfigMediaType = "application/vnd.devfileio.devfile.config.v2+json"
+	devfileMediaType       = "application/vnd.devfileio.devfile.layer.v1"
+	registryPath           = "/registry/stacks"
+	indexPath              = "/registry/index.json"
```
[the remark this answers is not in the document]
johnmcollier: Maybe instead of storing the paths in variables, we should just get the value from `DEVFILE_STACKS` and `DEVFILE_INDEX` env vars?

index-server/main.go
```diff
@@ -0,0 +1,151 @@
...
+						"error":  err.Error(),
+						"status": fmt.Sprintf("failed to pull the devfile of %s", name),
+					})
+				}
+				c.Data(http.StatusOK, http.DetectContentType(bytes), bytes)
+			}
+		}
+	})
+
+	router.StaticFile("/index", indexPath)
```
johnmcollier, line 96: Can we serve the index.json on `/` instead?

johnmcollier (changes requested):
Left some review comments.

Another suggestion I had is that since this is called `index-server`, does it make sense to move it under the `index/` folder in this repo?

e.g.
```
index/
   generator/
   server/
```

index-server/main.go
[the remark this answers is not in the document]
GeekArthur: Good point, we can make it more dynamic and get those data from `DEVFILE_STACKS` and `DEVFILE_INDEX` env vars, user only need to specify the env vars.

index-server/main.go
```diff
@@ -0,0 +1,151 @@
...
+		if err != nil {
+			log.Fatal(err.Error())
+		}
+
+		if resp.StatusCode == http.StatusOK {
+			isDone = true
+			log.Println("Registry is up and running")
+		}
+		log.Println("Waiting for registry to start...")
+		time.Sleep(time.Second)
```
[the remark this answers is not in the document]
GeekArthur: That's fair, I make it shorter.

index-server/main.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	// Load index file
+	bytes, err := ioutil.ReadFile(indexPath)
+	if err != nil {
+		log.Fatalf("failed to read index file: %v", err)
+	}
+
+	var index []indexSchema.Schema
```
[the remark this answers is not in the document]
GeekArthur: Yep, I add the ToDo notes here with corresponding open issues (eg. I believe index.json parsing is this issue: https://github.com/devfile/api/issues/223).

index-server/Dockerfile
GeekArthur: Sure, I can update that.

index-server/main.go
```diff
@@ -0,0 +1,151 @@
...
+		c.JSON(http.StatusOK, gin.H{
+			"message": "the server is up and running",
+		})
+	})
+
+	router.GET("/devfiles/:name", func(c *gin.Context) {
+		name := c.Param("name")
+		for _, devfileIndex := range index {
+			if devfileIndex.Name == name {
+				bytes, err := pullStackFromRegistry(devfileIndex)
```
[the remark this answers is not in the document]
GeekArthur: Originally I design the way like that is because it's easier to let consumers integrate (eg. odo already implemented HTTP request internally so odo just need to update the endpoint we provide), but since we decide to let consumers use the library we provide for integration, then I think it's fine we remove the endpoint `/devfiles/<name>` and let user browser the whole stacks by serving the whole stacks as static files.

I can give a try to see if Gin API `router.StaticFile()` can serve the whole stacks properly so that user can view the whole stacks instead of only devfile.

index-server/entrypoint.sh
GeekArthur: This is interesting, actually I had the same thought as yours and tried the code block above (starts index server process firstly at backgeound), but it failed, I went into the index container for deeper debugging, for some reasons the index server process was not in the index container and I can only saw the nginx server process. I couldn't figure out the reasons but probably it's due to docker doesn't have a really good support for the case where multiple services in one container, especially multiple services need communication. Probably entrypoint process must start firstly? 🤔

index-server/main.go
GeekArthur: Yep, I can do that.

index-server/entrypoint.sh
GeekArthur: `exec` command uses the existing bash process to run the command instead of creating new process and docker uses it as entrypoint process, from docker perspective, I am not sure if that results in another background process that we trigger in the same script gets ignored.

johnmcollier: Huh, interesting. In that case, let's stick with that you're doing.

GeekArthur:
> Left some review comments.
> 
> Another suggestion I had is that since this is called `index-server`, does it make sense to move it under the `index/` folder in this repo?
> 
> e.g.
> 
> ```
> index/
>    generator/
>    server/
> ```

I though about the same thing before, will do that.

GeekArthur:
@johnmcollier comments addressed, please review again, thanks!

johnmcollier (changes requested):
Every time the index server starts for the first time, it exits with an error:
```
^CJohns-MacBook-Pro-3:kubernetes johncollier$ oc get po
NAME                                                   READY   STATUS    RESTARTS   AGE
devfile-registry-8467d84c6d-zrwgw                      2/2     Running   1          4m20s
registry-operator-controller-manager-9b6b557b4-79hpz   2/2     Running   0          15m
```

I think it's happening because of the `log.Fatal` we do if the `http.Get` to the registry server returns an error. Probably the registry server isn't quite ready yet and isn't giving back an http response code?) 

I think it might be better if we don't exit when an `http.Get` returns an error, but rather keep trying up to a given timeout?

This is what I did in the registry operator when I needed to poll a given URL (see https://github.com/johnmcollier/registry-operator/blob/main/pkg/util/util.go#L25)
```
func WaitForServer(url string, timeout time.Duration) error {
	return wait.PollImmediate(time.Second, timeout, func() (bool, error) {
		tr := &http.Transport{
			TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
		}
		client := &http.Client{Transport: tr}
		resp, err := client.Get(url)
		if err != nil {
			return false, err
		}
		if resp.StatusCode == 200 {
			return true, nil
		}
		return false, nil
	})
}
```

I'd recommend setting the timeout to no more than 30 seconds, since any more and the Kube liveness probe will fail and kill the container. You may also want to change the `time.Second` in `wait.PollImmediate` to some multiple of `time.Millisecond`.

index/server/main.go
```diff
@@ -0,0 +1,142 @@
...
+	indexPath  = os.Getenv("DEVFILE_INDEX")
+)
+
+func main() {
+	// Wait until registry is up and running
+	isDone := false
+	for !isDone {
+		resp, err := http.Get(scheme + "://" + registryService)
+		if err != nil {
+			log.Fatal(err.Error())
```
johnmcollier, line 43: Looks like `Connection refused` errors from the registry server will also lead to `http.Get` returning an error, so I don't think we can safely do a log.Fatal here.

This is what I get if I change line 43 to log the error and continue:
```
2020/11/26 19:45:55 Get "http://localhost:5000": dial tcp [::1]:5000: connect: connection refused
```

GeekArthur: You are right, this shouldn't be a place to throw fatal error directly, I can make it fail until it reaches the timeout.

GeekArthur:
@johnmcollier Can you help review again? Thanks

index/server/main.go
```diff
@@ -36,19 +37,22 @@ var (
 
 func main() {
 	// Wait until registry is up and running
-	isDone := false
-	for !isDone {
+	err := wait.PollImmediate(time.Millisecond, time.Second*30, func() (bool, error) {
 		resp, err := http.Get(scheme + "://" + registryService)
 		if err != nil {
-			log.Fatal(err.Error())
+			return false, err
```
johnmcollier, line 43: I'm still hitting the crash on first time start up:
```
Johns-MacBook-Pro-3:registry-operator johncollier$ k logs -f -c devfile-registry-bootstrap test-devfile-registry-cffb7559b-nbspk
2020/12/01 17:59:45 Get "http://localhost:5000": dial tcp 127.0.0.1:5000: connect: connection refused
/entrypoint.sh: fg: line 19: job %1 not created under job control
```

Looks like this was an issue with the sample I provided you. It looks like `wait.PollImmediate` will return an error early if the function passed into it ever returns a non-nil error.

So I think we need to change this line to: `return false, nil`

johnmcollier (changes requested):

index/server/main.go
johnmcollier: Yup, confirmed that was the issue. I would change the `if err != nil` to:
```
if err != nil {
			log.Println(err.Error())
			return false, nil
		}
```

Log the error too so that it's not lost. We just don't want to return it.

GeekArthur: My original understanding was that only return `true, err` can cause the polling function throw error within timeout, but seems like from your testing it doesn't work like that. But interesting thing is my cluster doesn't throw the error even with return `false, err` (probably it depends on the time of registry is up and running).

My concern is that if we return `false, nil` and we don't return it, then the real registry failure won't bring the server down.

johnmcollier: Well, until `true, nil` is returned, the polling function will keep retrying up until the timeout is reached. So if there is something wrong, the timeout will be reached and an error will be returned.

GeekArthur: I check the code again, I that would be fine as we can bring the server down on pushing devfile stage instead of checking registry is up or not stage, not perfect, but still works fine, only delay couple of seconds I think.

GeekArthur: I am gonna make the changes as you suggest.

GeekArthur:
@johnmcollier changes are pushed, please review.

johnmcollier (approved):
LGTM

PR #14, opened 2020-12-03
Migrate registry build tools over from devfile/registry

johnmcollier:

Migrates the existing build scripts and Dockerfile from https://github.com/devfile/registry to this repository. No functional changes were made to the build tools, just a small change to allow the registry folder to be specified. Once this PR is merged, I'll remove the build tools from the registry repo.

The build tools are still relatively simplistic, but we can make further, centralized improvements to them in this repository now (under the `build-tools` folder), rather than on a per-registry basis.

N/A

**PR acceptance criteria**:
N/A

1. Git clone this repository and devfile/registry
2. cd to `build-tools` in this repository
3. run `./build.sh <path-to-cloned-devfile/registry>`

Build should succeed and produce a devfile index container image.

### DIFF
--- a/build-tools/build.sh
+++ b/build-tools/build.sh
@@ -9,12 +9,12 @@
 
 #!/bin/sh
 
-buildToolsFolder="$(basename "$(dirname "$0")")"
+buildToolsFolder="$(dirname "$0")"
 generatorFolder=$buildToolsFolder/../index/generator
 
 display_usage() { 
   echo "A devfile registry repository folder must be passed in as an argument" 
-  echo "usage: ./build.sh <path-to-registry-repository-folder>" 
+  echo "usage: build.sh <path-to-registry-repository-folder>" 
 } 
 
 # Check if a registry repository folder was passed in, if not, exit
@@ -24,11 +24,12 @@ if [ $# -ne 1 ]; then
   exit 1
 fi
 
+cd $generatorFolder
+
 # Build the index generator/validator
 echo "Building index-generator tool"
-cd $generatorFolder
 ./build.sh
-if [ ! $? -eq 0 ]; then
+if [ $? -ne 0 ]; then
   echo "Failed to build index-generator tool"
   exit 1
 fi
@@ -39,7 +40,7 @@ cd "$OLDPWD"
 # Run the index generator tool
 echo "Generate the devfile registry index"
 $generatorFolder/index-generator $registryFolder/stacks $registryFolder/index.json
-if [ ! $? -eq 0 ]; then
+if [ $? -ne 0 ]; then
   echo "Failed to build the devfile registry index"
   exit 1
 fi
@@ -48,7 +49,7 @@ echo -e "Successfully built the devfile registry index\n"
 # Build the Docker image containing the devfile stacks and index.json
 echo "Building the devfile registry index container"
 docker build -t devfile-index -f $buildToolsFolder/Dockerfile $registryFolder
-if [ ! $? -eq 0 ]; then
+if [ $? -ne 0 ]; then
   echo "Failed to build the devfile registry index container"
   exit 1
 fi
--- a/build-tools/build.sh
+++ b/build-tools/build.sh
@@ -1,0 +1,56 @@
+#
+# Copyright (c) 2020 Red Hat, Inc.
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+#!/bin/sh
+
+buildToolsFolder="$(basename "$(dirname "$0")")"
+generatorFolder=$buildToolsFolder/../index/generator
+
+display_usage() { 
+  echo "A devfile registry repository folder must be passed in as an argument" 
+  echo "usage: ./build.sh <path-to-registry-repository-folder>" 
+} 
+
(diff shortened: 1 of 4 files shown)


build-tools/build.sh
```diff
@@ -0,0 +1,56 @@
...
+if [ $# -ne 1 ]; then
+  display_usage
+  exit 1
+fi
+
+# Build the index generator/validator
+echo "Building index-generator tool"
+cd $generatorFolder
+./build.sh
+if [ ! $? -eq 0 ]; then
```
GeekArthur, line 31: Nit: can directly use `if [ $? -ne 0]; then`

build-tools/build.sh
```diff
@@ -0,0 +1,56 @@
...
+  exit 1
+fi
+echo -e "Successfully built the index-generator tool\n"
+
+cd "$OLDPWD"
+
+# Run the index generator tool
+echo "Generate the devfile registry index"
+$generatorFolder/index-generator $registryFolder/stacks $registryFolder/index.json
+if [ ! $? -eq 0 ]; then
```
GeekArthur, line 42: Nit: can directly use `if [ $? -ne 0]; then`

build-tools/build.sh
```diff
@@ -0,0 +1,56 @@
...
+  echo "Failed to build the devfile registry index"
+  exit 1
+fi
+echo -e "Successfully built the devfile registry index\n"
+
+# Build the Docker image containing the devfile stacks and index.json
+echo "Building the devfile registry index container"
+docker build -t devfile-index -f $buildToolsFolder/Dockerfile $registryFolder
+if [ ! $? -eq 0 ]; then
```
GeekArthur, line 51: Nit: can directly use `if [ $? -ne 0]; then`

build-tools/build.sh
```diff
@@ -0,0 +1,56 @@
...
+# This program and the accompanying materials are made
+# available under the terms of the Eclipse Public License 2.0
+# which is available at https://www.eclipse.org/legal/epl-2.0/
+#
+# SPDX-License-Identifier: EPL-2.0
+#
+
+#!/bin/sh
+
+buildToolsFolder="$(basename "$(dirname "$0")")"
```
GeekArthur, line 12: I think we can just use `"$(dirname "$0")"` then buildToolsFolder will store the absolute path of build tools folder instead of relative path of build tools folder, using absolute path can ensure user can run the build script anywhere. (relative path only support user run build script in registry-support repo).

build-tools/build.sh
```diff
@@ -0,0 +1,56 @@
...
+# Check if a registry repository folder was passed in, if not, exit
+registryFolder=$1
+if [ $# -ne 1 ]; then
+  display_usage
+  exit 1
+fi
+
+# Build the index generator/validator
+echo "Building index-generator tool"
+cd $generatorFolder
```
GeekArthur, line 29: I think we can specify the generator build script and generator binary path at the beginning of this script as global variable, then we don't need to switch directory back and forth by using `cd <path>` + `cd "$OLDPWD"` way, it's also better for readability and maintainability I think.

build-tools/build.sh
```diff
@@ -0,0 +1,56 @@
...
+#
+
+#!/bin/sh
+
+buildToolsFolder="$(basename "$(dirname "$0")")"
+generatorFolder=$buildToolsFolder/../index/generator
+
+display_usage() { 
+  echo "A devfile registry repository folder must be passed in as an argument" 
+  echo "usage: ./build.sh <path-to-registry-repository-folder>"
```
GeekArthur, line 17: IMO, probably we can remove the part of path specify: `./` as the script can be run anywhere, so the usage would be `echo "usage: build.sh <path-to-registry-repository-folder>" `

GeekArthur (changes requested):

build-tools/build.sh
johnmcollier: Good catch 👍

build-tools/build.sh
johnmcollier: So, I tried to do that initially when I wrote the script, but the index generator build doesn't support being run outside of its own directory:
```
Johns-MacBook-Pro-3:registry-support johncollier$ index/generator/build.sh 
build flag -mod=vendor only valid when using modules
```

build-tools/build.sh
johnmcollier: 👍

build-tools/build.sh
johnmcollier: Updated.

build-tools/build.sh
johnmcollier: Updated.

build-tools/build.sh
johnmcollier: Updated

build-tools/build.sh
johnmcollier: Updated.

build-tools/build.sh
johnmcollier: Updated.

johnmcollier:
@GeekArthur Thanks for the review. Review comments addressed, see my note about the index generator build.

build-tools/build.sh
GeekArthur: Yeah, I think so, we have to inside the package dir to build go module, then sounds like we have to move the same thing (cd <path> + cd "$OLDPWD") to index generator build.sh. Given that I think we can keep the existing code.

GeekArthur (approved):
Thanks for making the changes, LGTM!

PR #16, opened 2020-12-15
Update registry bootstrap to use resources struct

johnmcollier:

This PR updates the registry bootstrap code to push contents to the devfile registry based on the logic outlined in the [Devfile Registry Structure](https://github.com/devfile/api/blob/master/docs/proposals/registry/registry-structure.md) design doc. 

Fixes https://github.com/devfile/api/issues/226

1. Deploy registry operator
2. Create an instance of the registry and set `johncollier/devfile-index:latest` as the devfile index image
3. Registry should start successfully.


index/server/vendor/github.com/devfile/registry-support/index/generator/schema/schema.go
```diff
@@ -59,5 +59,6 @@ type Schema struct {
 	ProjectType     string            `yaml:"projectType,omitempty" json:"projectType,omitempty"`
 	Language        string            `yaml:"language,omitempty" json:"language,omitempty"`
 	Links           map[string]string `yaml:"links,omitempty" json:"links,omitempty"`
+	Resources       []string          `yaml:"resources,omitempty" json:"resources,omitempty"`
```
GeekArthur, line 62: Can we add the definition and example for the new field `Resources`?

index/server/main.go
```diff
@@ -23,13 +24,25 @@ import (
...
+	devfileSVGLogoMediaType = "image/svg+xml"
+	devfilePNGLogoMediaType = "image/png"
+	devfileArchiveMediaType = "application/x-tar"
+	scheme                  = "http"
+	registryService         = "localhost:5000"
 )
 
+var mediaTypeMapping = map[string]string{
+	"devfile.yaml": devfileMediaType,
+	"vsx":          devfileVSXMediaType,
```
GeekArthur, line 40: Nit: we can initialize consts for the resources as well so that we can use the consts here for map initialization and the switch block below (eg. `case devfileName, "logo.svg", "logo.png", "archive.tar":`)

GeekArthur (changes requested):

index/server/vendor/github.com/devfile/registry-support/index/generator/schema/schema.go
johnmcollier: This is just the vendored changes being pulled in from the generator, it needs to be updated there.

johnmcollier: Since the generator and index server live in the same repo, I added a commit to include examples in the index schema.

GeekArthur: I see 👍

GeekArthur (approved):
LGTM!

PR #19, opened 2021-01-11
Enable metrics on the registry index server

johnmcollier:

Adds a /metrics endpoint via the Prometheus Go client library to the index server container, running on a separate port (as we don't want the endpoint exposed publicly). As the registry index server is still relatively simple, I've opted not to expose _additional_ metrics beyond what the library provides for the time being (especially as the OCI container provides additional metrics), but it's something we may want to revisit as expand the index server.

Fixes https://github.com/devfile/api/issues/283

1) Deploy a devfile registry using the registry operator, specify `johncollier/devfile-index:metrics` as the `devfileIndexImage`
2) Run `kubectl exec <devfile-registry-pod> -c devfile-registry-metadata -- curl localhost:7001`


GeekArthur (approved):
LGTM, thanks for enabling the metrics for index server.

PR #20, opened 2021-01-13
Fix GitHub action that builds the images

johnmcollier:

The GitHub actions that we switched to now require us to set `path` for the Docker build context for each image that we're building.


johnmcollier:
Test action can be found here: https://github.com/johnmcollier/registry-support/runs/1697184658?check_suite_focus=true

GeekArthur (approved):
LGTM!

PR #22, opened 2021-01-22
Generate the index.json after creating archive.tar

johnmcollier:

This PR changes the registry build script so that the index.json is generated *after* creating any necessary archive.tar files so that the archive properly gets picked up in the index.json.

**PR acceptance criteria**:

N/A


GeekArthur (approved):
Good catch, it looks good to me!

PR #23, opened 2021-01-26
Implement registry library

GeekArthur:

This PR implements the registry library to allow consumers to interact with the registry, currently the registry library supports:
1. Get the devfile that belongs to a given stack
2. Get the devfile + the containing stack resource files of a given stack

Fixes https://github.com/devfile/api/issues/224

**PR acceptance criteria**:

- [x] Test (WIP) 

- [x] Documentation (WIP)


registry-library/library/library.go
```diff
@@ -0,0 +1,190 @@
...
+	fmt.Fprintln(w, "Name", "\t", "Description", "\t")
+	for _, stackIndex := range registryIndex {
+		fmt.Fprintln(w, stackIndex.Name, "\t", stackIndex.Description, "\t")
+	}
+	w.Flush()
+	return nil
+}
+
+// PullStackFromRegistry pulls stack from registry with allowed media types
+func PullStackFromRegistry(registry string, stack string, allowedMediaTypes []string) error {
```
johnmcollier, line 91: To be honest, I'm not crazy about the `PullStackFromRegistry` function requiring users to specify the media types. Most consumers will only ever need to pull all of the files in the stack

 It should always use `devfileAllMediaTypes` and maybe we should have a separate function that allows users/devs to specify the media types they want pulled

registry-library/library/library.go
```diff
@@ -0,0 +1,190 @@
...
+
+// ListRegistryStacks lists the stacks of the registry
+func ListRegistryStacks(registry string) error {
+	// Get the registry index
+	registryIndex, err := GetRegistryIndex(registry)
+	if err != nil {
+		return err
+	}
+
+	w := tabwriter.NewWriter(os.Stdout, 5, 2, 3, ' ', tabwriter.TabIndent)
```
johnmcollier, line 81: I don't think we should directly print out to stdout here. We should leave any printing or formatting to whatever is calling this function, as we're not sure what format they want it displayed in, how it's being displayed, or if there's additional/less information they need displayed.

If we do need this function, we should rename it to `PrintRegistryStacks` and/or return a formatted string rather than printing to stdout.

registry-library/library/library.go
```diff
@@ -0,0 +1,190 @@
...
+	responseHeaderTimeout = 30 * time.Second // responseHeaderTimeout is the timeout to retrieve the server's response headers
+)
+
+var (
+	stacksPath = os.Getenv("DEVFILE_STACKS")
+	indexPath  = os.Getenv("DEVFILE_INDEX")
+)
+
+// GetRegistryIndex gets the index of the specified registry
+func GetRegistryIndex(registry string) ([]indexSchema.Schema, error) {
```
johnmcollier, line 44: I think we should call this either `GetRegistryStacks` (or to `ListRegistryStacks` and change the function below to `PrintRegistryStacks`) instead as it's more clear what it does than `GetRegistryIndex`.

johnmcollier (changes requested):
I left some review comments

Some non-specific comments that I had as well:
1) We should look at testing on the OpenShift CI against a devfile registry
2) Have we considered packaging this as part of the devfile library rather than having a separate Go library just for the registry? My thinking is that anything consuming this library will also need to consume the devfile library, and having both libraries together makes things more consistent. CC @elsony

elsony:
> 2. Have we considered packaging this as part of the devfile library rather than having a separate Go library just for the registry? My thinking is that anything consuming this library will also need to consume the devfile library, and having both libraries together makes things more consistent. CC @elsony

It may not be true that anyone who accesses the registry will need to access the devfile library. Take the Che scenario, my understanding is that they have multiple components who need to access the registry and some of them just need to do the list but do not need to access the actual devfile itself. Having said that, I am open to discuss on if we want to lump them under the same library or not. I just added an item to the next team meeting to discuss.

johnmcollier:
@elsony Sounds good 👍

registry-library/library/library.go
```diff
@@ -0,0 +1,190 @@
...
+			break
+		}
+	}
+	if !exist {
+		return fmt.Errorf("Stack %s does not exist in the registry %s", stack, registry)
+	}
+
+	// Pull stack initialization
+	ctx := context.Background()
+	resolver := docker.NewResolver(docker.ResolverOptions{PlainHTTP: true})
```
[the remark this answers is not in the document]
GeekArthur: That's true, will make the scheme dynamic. Dose our registry already support the HTTPS? I didn't try and I am not very sure the configuration. Probably enable tls when deploying the registry by using the operation and set up some secrets?

registry-library/cmd/root.go
```diff
@@ -0,0 +1,135 @@
...
+	shortDesc = "Commands to interact with devfile registry"
+	longDesc  = "Commands to interact with devfile registry"
+)
+
+var (
+	registry             = os.Getenv("REGISTRY")
+	cfgFile              string
+	allResources         bool
+	devfileMediaType     = []string{library.DevfileMediaType}
+	devfileAllMediaTypes = []string{library.DevfileMediaType, library.DevfilePNGLogoMediaType, library.DevfileSVGLogoMediaType, library.DevfileVSXMediaType, library.DevfileArchiveMediaType}
```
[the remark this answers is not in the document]
GeekArthur: Yeah, I can move them to the library package.

registry-library/LICENSE
```diff
@@ -0,0 +1,202 @@
+
+                                 Apache License
```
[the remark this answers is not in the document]
GeekArthur: This is generated by cobra CLI automatically, will update all the licenses to EPL.

registry-library/build.sh
```diff
@@ -0,0 +1,2 @@
+#!/bin/sh
+CGO_ENABLED=0 go build -mod=vendor -o registry .
```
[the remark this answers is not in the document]
GeekArthur: Good catch, will update the GitHub CI.

registry-library/library/library.go
GeekArthur: Given golang doesn't support optional argument for function, instead of creating separate function to support the customized allowed media types, we may use the struct to wrap all the function arguments then pass the struct to the function, by default the `allowedMediaTypes` is `devfileAllMediaTypes` if the `allowedMediaTypes` is not specified in the struct.

registry-library/library/library.go
GeekArthur: Agree with you, I just create this function for testing the registry list and this function is only used by the CLI, I will add another generic in the middle.

registry-library/library/library.go
johnmcollier: That still seems confusing to people using the registry library though

Why not just do something like this, and then code isn't really duplicated?
```
func PullStackByMediaTypesFromRegistry(registry string, allowedMediaTypes []string) error)
func PullStackFromRegistry(registry string) {
  return PullStackByMediaTypesFromRegistry(registry, devfileAllMediaTypes)
}
```

registry-library/library/library.go
GeekArthur: I am not sure if `GetRegistryStacks/ListRegistryStacks` is more clear to explain the functionality of this function as this function is only return the content of the registry index, maybe `GetRegistryIndex` is already self-explanatory.

johnmcollier: It's a bit of a nit and yes, that name is self explanatory, but I'm thinking of the library adopters though or anyone trying to get the list of stacks from a registry. To get the list of stacks, this function (GetRegistryIndex) index has to be called. If someone's not entirely familiar with what an index.json or index entry is, they may not know that this function returns the list of stacks.

johnmcollier: Like I said though, it's a nit, so if you prefer `GetRegistryIndex` I'm fine with that.

registry-library/library/library.go
GeekArthur: That's fine for me, I can do the following function signatures, also add the missing `stack` argument

```
// Public function so that consumer can directly call it directly by case
func PullStackByMediaTypesFromRegistry(registry string, stack string, allowedMediaTypes []string) error {
    // As above...
} 

func PullStackFromRegistry(registry string, stack string) error {
  return PullStackByMediaTypesFromRegistry(registry, stack, devfileAllMediaTypes)
}
```

registry-library/library/library.go
GeekArthur: I see, if that's the case, I can rename the function to `GetRegistryStacks/ListRegistryStacks`, and explain the functionality (eg. related to index.json/index) more in the comment so that it's more friendly for consumer who is not familiar with registry structure and includes the functionality details as well.

registry-library/library/library.go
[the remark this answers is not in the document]
johnmcollier: If you need an env to test on, for OpenShift it's pretty straightforward:
1. Install operator
2. Install registry with cluster's certificates:
```
$ cat <<EOF | oc apply -f -
apiVersion: registry.devfile.io/v1alpha1
kind: DevfileRegistry
metadata:
  name: devfile-registry
spec:
  devfileIndexImage: quay.io/devfile/devfile-index:next
EOF
```

(TLS is enabled by default on OpenShift, so the above yaml should get you a registry with HTTPS)

GeekArthur:
> I left some review comments
> 
> Some non-specific comments that I had as well:
> 
>     1. We should look at testing on the OpenShift CI against a devfile registry (post-merging)
I was thinking about the same, we may also need to look at the pre-merging case, like provision a devfile registry so that registry library can be tested.

>     2. Have we considered packaging this as part of the devfile library rather than having a separate Go library just for the registry? My thinking is that anything consuming this library will also need to consume the devfile library, and having both libraries together makes things more consistent. CC @elsony
IMO, it depends on if some of existing consumers _only_ need to access the registry. If yes we should separate, but if no we can merge the registry library with devfile library. We need to figure out the use cases from consumers perspective.

johnmcollier:
> I was thinking about the same, we may also need to look at the pre-merging case, like provision a devfile registry so that registry library can be tested.

For that, I just meant it can be looked at after merging this PR, as we'll need to work with QE to get the tests running on openshift CI

GeekArthur:
@johnmcollier Thanks for reviewing the PR, comments addressed, please review again.

johnmcollier (approved):
LGTM

PR #26, opened 2021-02-04
Add pushing devfile-index image step in github CI

GeekArthur:

This PR adds pushing devfile-index image step in github CI so that both devfile-index-base image change and build tools change can trigger the build and push of devfile-index image.

### DIFF
--- a/.github/workflows/pushimage-next.yaml
+++ b/.github/workflows/pushimage-next.yaml
@@ -1,11 +1,12 @@
 #
-# Copyright (c) 2020 Red Hat, Inc.
+# Copyright (c) 2021 Red Hat, Inc.
 # This program and the accompanying materials are made
 # available under the terms of the Eclipse Public License 2.0
 # which is available at https://www.eclipse.org/legal/epl-2.0/
 #
 # SPDX-License-Identifier: EPL-2.0
 #
+
 name: Next Dockerimage
 
 on:
@@ -13,13 +14,18 @@ on:
     branches: [ master ]
 
 jobs:
-
   indexServerBuild:
     runs-on: ubuntu-latest
     steps:
-    - name: Checkout devworkspace-operator source code
+    - name: Check out registry support source code
       uses: actions/checkout@v2
-    - name: Docker Build & Push
+    - name: Check out registry source code
+      uses: actions/checkout@v2
+      with:
+        repository: devfile/registry
+        persist-credentials: false
+        path: registry
+    - name: Build and push devfile-index-base docker image
       uses: docker/build-push-action@v1.1.0
       with:
         path: ./index/server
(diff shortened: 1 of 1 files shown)


.github/workflows/pushimage-next.yaml
```diff
@@ -1,25 +1,31 @@
 #
-# Copyright (c) 2020 Red Hat, Inc.
+# Copyright (c) 2021 Red Hat, Inc.
```
johnmcollier, line 2: Nit: Should be 2020-2021

GeekArthur: Yeah, you are right, will update

johnmcollier (commented):
So this all looks fine to me, but I'm wondering if this is the approach we want to take here. This means that for each "devfile registry" repository we manage (so both the public and product registries), we'll need to duplicate its build & push steps in this repo's CI workflows as well.

What about using "repository dispatches" in this repo to trigger a build in the other repos? Seems there's a convenient GitHub action to do so, called [repository-dispatch](https://github.com/peter-evans/repository-dispatch). Something like the following:

In this repo's pushimage-next.yaml:
```
jobs:
  build:
    # Main workflow job that builds, tests, etc.

  dispatch:
    needs: build
    strategy:
      matrix:
        repo: ['devfile/registry', 'openshift/some-product-registry', 'another/registry']
    runs-on: ubuntu-latest
    steps:
      - name: Repository Dispatch
        uses: peter-evans/repository-dispatch@v1
        with:
          token: ${{ secrets.REPO_ACCESS_TOKEN }}
          repository: ${{ matrix.repo }}
          event-type: my-event
```

Then in `pushimage-next.yaml` in the devfile/registry repository change:
```

on:
  push:
    branches: [ master ]
```
to:
```
on:
  push:
    branches: [ master ]
  repository_dispatch:
    types: [my-event]
```

GeekArthur:
> So this all looks fine to me, but I'm wondering if this is the approach we want to take here. This means that for each "devfile registry" repository we manage (so both the public and product registries), we'll need to duplicate its build & push steps in this repo's CI workflows as well.
> 
> What about using "repository dispatches" in this repo to trigger a build in the other repos? Seems there's a convenient GitHub action to do so, called [repository-dispatch](https://github.com/peter-evans/repository-dispatch). Something like the following:
> 
> In this repo's pushimage-next.yaml:
> 
> ```
> jobs:
>   build:
>     # Main workflow job that builds, tests, etc.
> 
>   dispatch:
>     needs: build
>     strategy:
>       matrix:
>         repo: ['devfile/registry', 'openshift/some-product-registry', 'another/registry']
>     runs-on: ubuntu-latest
>     steps:
>       - name: Repository Dispatch
>         uses: peter-evans/repository-dispatch@v1
>         with:
>           token: ${{ secrets.REPO_ACCESS_TOKEN }}
>           repository: ${{ matrix.repo }}
>           event-type: my-event
> ```
> 
> Then in `pushimage-next.yaml` in the devfile/registry repository change:
> 
> ```
> 
> on:
>   push:
>     branches: [ master ]
> ```
> 
> to:
> 
> ```
> on:
>   push:
>     branches: [ master ]
>   repository_dispatch:
>     types: [my-event]
> ```

That's doable, can you help create a [personal access token](https://docs.github.com/en/github/authenticating-to-github/creating-a-personal-access-token) with `repo` scope for registry repo, I think you may get the admin access to do that when you set up the secrets for quay. Please let me know the secret name after you done that, I think we can use the name `REPO_ACCESS_TOKEN` for all kinds of registry repo to make the name consistent.

johnmcollier:
@GeekArthur I've set the `REPO_ACCESS_TOKEN` secret in this repo to be a PAT

GeekArthur:
@johnmcollier Can you please review again? I am gonna create another in registry repo and let you review.

johnmcollier (approved):

PR #27, opened 2021-02-09
Implement a metric to measure latency on the index server

johnmcollier:

This PR implements a new prometheus metric, `index_http_request_duration_seconds` to measure the latency on the `/` and `/index` endpoints.

As it is right now, we don't need similar metrics on the `/stacks` or `/devfile/<name>` endpoint.

Fixes https://github.com/devfile/api/issues/336

### DIFF
--- a/index/server/main.go
+++ b/index/server/main.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"github.com/prometheus/client_golang/prometheus"
 	"io/ioutil"
 	"log"
 	"net/http"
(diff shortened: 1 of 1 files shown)


GeekArthur (approved):
The code looks good to me, wonder do we use `histogram_quantile` function somewhere to measure the quantile of a histogram?

johnmcollier:
@GeekArthur That will be done cluster-side on the cluster hosting the devfile registry

GeekArthur:
> @GeekArthur That will be done cluster-side on the cluster hosting the devfile registry

I see, thanks for clarifying that.

PR #28, opened 2021-02-10
Index schema validator

GeekArthur:

This PR implements the following features:
1. Add index schema validator to index generator
2. Generate index.json from devfile.yaml instead of meta.yaml as we already merged the meta.yaml to devfile.yaml

Fixes https://github.com/devfile/api/issues/222

**PR acceptance criteria**:

- [x] Test (WIP) 

- [x] Documentation (WIP)

Run `build.sh` to build the index generator CLI then test the CLI with specific registry.


index/generator/library/library.go
```diff
@@ -74,3 +92,17 @@ func CreateIndexFile(index []schema.Schema, indexFilePath string) error {
 
 	return nil
 }
+
+func validateIndexComponent(indexComponent schema.Schema) error {
```
johnmcollier, line 96: Do we really need to validate the index.json in the index generator itself? I would assume that any index.json it spits out is valid, and we shouldn't need to validate that.

GeekArthur: IMO, we should at least validate the following things for index.json generation, otherwise the following operations (eg. list stacks, pull down stack) that are using the index.json will be blocked due to missing fields in index.json
- `indexComponent.Name` is not empty so that user can pull down stack resources by choosing the stack name  
- `indexComponent.Links` is not empty so that at least user can pull down the stack devfile (the devfile link is necessary for pulling down the stack devfile)
- `indexComponent.Resources` is not empty because at least the devfile should exist in resources field

johnmcollier: That's a fair point actually, and it can't hurt to do I suppose.

johnmcollier (approved):

PR #29, opened 2021-02-16
(Hosted Registry) Add configuration for number of registry replicas and remove routes

johnmcollier:
This PR adds a couple required changes to deploy the hosted registry on the managed OpenShift Dedicated environment:
- Adds a configuration in the OpenShift template to allow the number of devfile registry replicas to be configured. 
- Removes routes from the main `devfile-registry.yaml` OpenShift template. 
    - This is because OpenShift routes will be configured and managed through a separate process.
    - I've included a `route.yaml` with the routes for debugging purposes.

N/A

**PR acceptance criteria**:

- [ ] Test (WIP) 

- [ ] Documentation (WIP)

1. Log in to an OpenShift cluster
2. Run `REPLICAS=3 oc new-app -f deploy/hosted-registry/
3. The devfile registry should get deployed and there should be 3 pods deployed.

### DIFF
--- a/deploy/hosted-registry/devfile-registry.yaml
+++ b/deploy/hosted-registry/devfile-registry.yaml
@@ -19,7 +19,7 @@ objects:
       app: devfile-registry
     name: devfile-registry
   spec:
-    replicas: 1
+    replicas: ${{REPLICAS}}
     selector:
       matchLabels:
         app: devfile-registry
@@ -217,3 +217,7 @@ parameters:
   value: Always
   displayName: Devfile registry image pull policy
   description: Always pull by default. Can be IfNotPresent
+- name: REPLICAS
+  value: 1
+  displayName: Devfile registry replicas
+  description: The number of replicas for the hosted devfile registry service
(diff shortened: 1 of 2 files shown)


deploy/hosted-registry/devfile-registry.yaml
```diff
@@ -19,7 +19,7 @@ objects:
       app: devfile-registry
     name: devfile-registry
   spec:
-    replicas: 1
+    replicas: ${{REPLICAS}}
```
GeekArthur, line 22: I am using zsh, seems like `${{REPLICAS}}` doesn't work for zsh to interpret the env var, should we use `${REPLICAS}` instead? It should work for all shells.

GeekArthur (changes requested):

GeekArthur:
@johnmcollier The PR looks good to me, only leave one comment above.

deploy/hosted-registry/devfile-registry.yaml
johnmcollier: Gah, lemme try zsh, still on bash

johnmcollier: Updated to use single curly brackets, good find Jingfu 👍

GeekArthur (approved):
LGTM!

PR #30, opened 2021-02-22
Fix replicas parameter in OpenShift template

johnmcollier:
This PR fixes the `REPLICAS` parameter when deploying the hosted devfile registry through the app-sre CICD pipelines. The issue was that the `values` field for OpenShift parameters expects a string and not an integer, so quotes around the default value for `REPLICAS` was needed, as well as double curly brackets around the parameter when referencing it.

N/A


elsony (approved):
/lgtm

PR #31, opened 2021-02-23
Add label for registry service

johnmcollier:
Adds a `app: devfile-registry` label to the hosted registry service, so that Prometheus can find and scrape its metrics endpoints


GeekArthur (approved):
LGTM!

PR #38, opened 2021-03-31
Add OpenShift template for devfile.io redirector

johnmcollier:

Adds an OpenShift template for properly redirecting devfile.io to docs.devfile.io. Required to fix redirect that was broken when we moved to the AppSRE AWS nameservers.


GeekArthur (approved):
LGTM

PR #39, opened 2021-03-31
Allow replicas to be configued for redirector

johnmcollier:
AppSRE suggested we set the replicas to 3 at a minimum. So making the replicas field and parameter and defaulting it to 3.


GeekArthur (approved):
LGTM

PR #44, opened 2021-04-22
Split registry build tools into two scripts

johnmcollier:
Splits the registry build tools into two scripts, `build.sh` and `build_image.sh`
- `build.sh` runs the index generator and creates an archive.tar for each stack (if needed), and outputs to a given output directory
- `build_image.sh` calls `build.sh`, then builds the stack artifacts into the devfile index image


build-tools/build.sh
```diff
@@ -16,74 +16,106 @@ buildToolsFolder="$(dirname "$0")"
...
+
+  # If the speicifed output folder is not empty, exit.
+  if [ ! -z "$(ls -A $outputFolder)" ]; then
+    echo "The specified destination folder is not empty. Please specify an empty folder."
+    display_usage
+    exit 1
+  fi
+}
+
+# Check if a registry repository folder was passed in, if not, exit
```
GeekArthur, line 105: Nit: Should the comment be `Check if a registry repository folder and a output folder were passed in, if not, exit`?

GeekArthur:
@johnmcollier The PR looks good to me, I just left one nit comment.

build-tools/build.sh
johnmcollier: Yeah, true. I'll update it.

johnmcollier:
@GeekArthur Thanks, updated.

GeekArthur (approved):
LGTM, thanks for making the changes.

Probably you need to rebase my PR: https://github.com/devfile/registry-support/pull/45
The key point is the change from `$generatorFolder/index-generator $outputFolder/stacks $outputFolder/index.json` to `$generatorFolder/index-generator $outputFolder $outputFolder/index.json` as we need to pass the registry root dir to the new index generator to support stacks/samples from the registry and extraDevfileEntries.yaml.

PR #47, opened 2021-04-27
Exit with proper code for build script

johnmcollier:
Typo in the cleanup_exit_function in the registry build scripts. Should be `exit $1` instead of `exit 1`


GeekArthur (approved):
LGTM

PR #51, opened 2021-04-29
Update index type from string to enum

maysunfaisal:
Signed-off-by: Maysun J Faisal <maysunaneek@gmail.com>

Index Schema

Updating the index schema type from string to enum, brought up here in this PR https://github.com/openshift/console/pull/8687#discussion_r623406909 and thought this was a good small change to make

I've also searched for `"sample"`  and `"stack"` strings and replaced them

N/A

**PR acceptance criteria**:

- [x] Test (Updated) 

- [ ] Documentation (WIP)

updated the tests, so pls run the go tests

### DIFF
--- a/index/generator/schema/schema.go
+++ b/index/generator/schema/schema.go
@@ -57,7 +57,7 @@ version: string - The stack version
 attributes: map[string]apiext.JSON - Map of implementation-dependant free-form YAML attributes
 displayName: string - The display name of devfile
 description: string - The description of devfile
-type: string - The type of the devfile, currently support stack and sample
+type: DevfileType - The type of the devfile, currently supports stack and sample
 tags: string[] - The tags associated to devfile
 icon: string - The devfile icon
 globalMemoryLimit: string - The devfile global memory limit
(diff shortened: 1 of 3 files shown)


jhadvig (approved):
@maysunfaisal thanks for makingthis clear 👍 
LGTM

index/generator/schema/schema.go
```diff
@@ -88,6 +88,17 @@ type Schema struct {
 	Git               *Git                   `yaml:"git,omitempty" json:"git,omitempty"`
 }
 
+// IndexType describes the type of index entry
+type IndexType string
```
GeekArthur, line 92: I think the type should of `DevfileType` as currently we support sample devfile and stack devfile. We also need to update the comment to make it more clear as the index entry doesn't have type.

I assume the type will be used on dev console side and that's why we export it, otherwise we can make the type local.

index/generator/schema/schema.go
```diff
@@ -88,6 +88,17 @@ type Schema struct {
...
+
+const (
+	// SampleIndexType represents a sample index
+	SampleIndexType IndexType = "sample"
```
GeekArthur, line 96: Update `SampleIndexType` to `SampleDevfileType` and the comment.

index/generator/schema/schema.go
```diff
@@ -88,6 +88,17 @@ type Schema struct {
...
+
+	// StackIndexType represents a stack index
+	StackIndexType IndexType = "stack"
```
GeekArthur, line 99: Update `StackIndexType` to `StackDevfileType` and the comment.

GeekArthur (changes requested):

GeekArthur:
@maysunfaisal I left some comments, I will come back to try this PR to see if it works well.

index/generator/schema/schema.go
maysunfaisal: updated!

index/generator/schema/schema.go
maysunfaisal: updated!

index/generator/schema/schema.go
maysunfaisal: Makes sense, updated!

Schema structs should remain exported because it allows access to others much like devfile/api schema structs.

GeekArthur (approved):
LGTM, thanks for making the changes.

PR #52, opened 2021-05-05
Registry REST API for sample support

GeekArthur:

Index server

This PR implements the registry REST API to filter stack vs sample

Relates issues:
https://github.com/devfile/api/issues/279
https://github.com/devfile/api/issues/403
https://github.com/devfile/api/issues/409

**PR acceptance criteria**:

- [x] Test (WIP) 

- [x] Documentation (WIP)

Deploy the registry then try the following endpoints:
- /index
- /index/sample
- /index/stack


index/server/Dockerfile
```diff
@@ -38,6 +38,8 @@ USER www-data
 # Set env vars for the locations of the devfile stacks and index.json
 ENV DEVFILE_STACKS /stacks
 ENV DEVFILE_INDEX /index.json
+ENV DEVFILE_SAMPLE_INDEX /home/www-data/sample_index.json
```
johnmcollier, line 41: Why do we put the sample index files under `/home/www-data`?

index/server/main.go
```diff
@@ -111,14 +114,31 @@ func main() {
...
+		}
+
 		if len(devfileIndex.Resources) != 0 {
 			err := pushStackToRegistry(devfileIndex)
 			if err != nil {
 				log.Fatal(err.Error())
 			}
 		}
 	}
+	err = indexLibrary.CreateIndexFile(sampleIndex, sampleIndexPath)
```
johnmcollier, line 134: Rather than re-generating parts of the index.json for both stacks and samples, why not just do something like the following for each endpoint:

1) Parse the index.json file from disk
2) If samples are requested, return only the entries samples. If stacks are requested, return those.

index/server/main.go
```diff
@@ -127,6 +147,21 @@ func main() {
 	router.GET("/index", serveDevfileIndex)
 	router.GET("/index.json", serveDevfileIndex)
 
+	router.GET("/index/:type", func(c *gin.Context) {
```
johnmcollier, line 150: This approach all work fine, but by keeping the `/index` endpoint serving _all_ stack types, it means that odo (which is currently consuming the registry library) needs to be updated to a newer version of the registry library that calls one of these endpoints instead. It also means that we can't update the hosted registry to add the samples until odo is using the stacks-only endpoint.

Would it be better to have `/index` just return stacks and have a separate endpoint for returning stacks and samples?

johnmcollier (changes requested):
Left some comments.

How has this been tested?

index/server/main.go
johnmcollier: Looks good to me. I’m fine with either approach, so let’s see what Elson suggests.

GeekArthur: Discussed this in today's cabal call, I will implement the new REST API design above: https://github.com/devfile/registry-support/pull/52#discussion_r629482807

GeekArthur: Updated

GeekArthur:
@johnmcollier Comments addressed, please review again.

index/server/main.go
```diff
@@ -111,14 +114,31 @@ func main() {
 	}
 
 	// Before starting the server, push the devfile artifacts to the registry
+	// Build sample_index.json and stack_index.json given index.json
+	var sampleIndex []indexSchema.Schema
+	var stackIndex []indexSchema.Schema
 	for _, devfileIndex := range index {
+		if devfileIndex.Type == "sample" {
```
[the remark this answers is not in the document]
GeekArthur: Updated

GeekArthur:
Comments addressed, @johnmcollier @maysunfaisal please review again.

johnmcollier (approved):

PR #53, opened 2021-05-13
Registry library sample support

GeekArthur:

Registry library

Related issues:
https://github.com/devfile/api/issues/279
https://github.com/devfile/api/issues/403
https://github.com/devfile/api/issues/409

**PR acceptance criteria**:

- [x] Test (WIP) 

- [x] Documentation (WIP)

Build the registry CLI then run `registry list --type sample` to list samples from the registry


registry-library/library/library.go
```diff
@@ -55,16 +55,22 @@ var (
...
 )
 
-// GetRegistryStacks returns the list of stacks, more specifically
-// it gets the content of the index (index.json) of the specified registry
-// for listing the stacks
-func GetRegistryStacks(registry string) ([]indexSchema.Schema, error) {
+// GetRegistry returns the list of stacks and/or samples, more specifically
+// it gets the stacks and/or samples content of the index of the specified registry
+// for listing the stacks and/or samples
+func GetRegistry(registry string, devfileType indexSchema.DevfileType) ([]indexSchema.Schema, error) {
```
johnmcollier, line 61: Personally, I don't like the name `GetRegistry` for this function. I understand why it was renamed, but personally I think it's too ambiguous.

`GetRegistryIndex` maybe? Any other ideas?

registry-library/library/library.go
```diff
@@ -55,16 +55,22 @@ var (
...
 	urlObj, err := url.Parse(registry)
 	if err != nil {
 		return nil, err
 	}
-	urlObj.Path = path.Join(urlObj.Path, "index")
+	if devfileType == indexSchema.StackDevfileType {
+		urlObj.Path = path.Join(urlObj.Path, "index")
+	} else if devfileType == indexSchema.SampleDevfileType {
+		urlObj.Path = path.Join(urlObj.Path, "index", "sample")
+	} else {
```
johnmcollier, line 71: Will this line work when the only types for `indexSchema.DevfileType` are stack or sample? Shouldn't we have one for both?

johnmcollier (changes requested):

registry-library/library/library.go
johnmcollier: We almost need a name for something that can refer to either stacks _or_ samples, but I don't know what that would be 🤔

registry-library/library/library.go
maysunfaisal: Would like to point out that I dont think the Index schema devfile type is mandatory.. We should probably also enforce that IMO

johnmcollier: Yeah, agreed.

registry-library/library/library.go
GeekArthur: I think `GetRegistryIndex` is more specific and it looks good to me.

registry-library/library/library.go
GeekArthur: > Will this line to get both stacks and samples work when the only types for `indexSchema.DevfileType` are stack or sample?

Yeah, it works. The current behavior is:
If user specify the devfile type -> return the list with the devfile type
If user doesn't specify the devfile type -> return the list contains all devfile types

> Would like to point out that I dont think the Index schema devfile type is mandatory.. We should probably also enforce that IMO

Currently user can pass `""` to the function, that means no devfile type is specified. But I think you mean to make the function argument `devfileType` optional?

maysunfaisal: No my argument is, index should always have `sample`/`stack`, it should not allow `""`

GeekArthur: Ok, if that's the case, we can one more devfile type called `all`, how do you think?

registry-library/library/library.go
GeekArthur: Updated

registry-library/library/library.go
GeekArthur: Discussed with Maysun over the slack, we decide to use variadic functions like `func GetRegistry(registry string, devfileType ...indexSchema.DevfileType) ([]indexSchema.Schema, error)` instead of adding new devfile type `all` because devfile type will be used and displayed in index.json and other metadata file, devfile type `all` doesn't make sense in those files.

GeekArthur: Updated

GeekArthur:
@johnmcollier @maysunfaisal Comments addressed, please review again, thanks.

johnmcollier (approved):
Changes look fine to me

maysunfaisal:
@GeekArthur lgtm, I would just put a simple test to make sure we're getting it what we want with that func, since other repos are consuming it. And probably a README about it since other repos are going to use it.

GeekArthur:
@maysunfaisal Yes, you can use the registry CLI to test the registry library. Currently I am working on registry library docs as part of docs restructure work, but I can open another for this repo's README.

PR #59, opened 2021-06-01
Enable the devfile registry integrations tests

johnmcollier:

This PR enables the devfile registry integration tests previously delivered on this repo to run on pull requests and after merging to master. In all it does the following:
- Adds a mock devfile registry under `test/registry`
    - Contains a couple simple devfiles and an `extraDevfileEntries.yaml`
    - A mock devfile registry is used rather than [devfile/registry](https://github.com/devfile/registry) to allow new features to be tested that haven't made it into the public registry repository yet.
- Adds a test dockerfile to build the mock registry content, using the registry build tools
- Adds CI script for building all of the components of this repository and to deploy on Minikube.
    - **Note:** Due to some trickery that I'm doing to avoid having to push the PR test images to a container registry, this script can only run on Linux **with** Minikube
    - It does the following, in order:
        1) Builds the index server base image
        2) Builds the mock devfile registry container image, which is based on the image built in step 1
        3) Deploys the devfile registry container image built in step 2, using the devfile registry helm chart
        4) Builds and runs the devfile registry integration test image
- Makes some small modifications to the devfile registry Helm chart to reduce its memory footprint (gives us more leeway when running on the GitHub Actions workflow)   

Fixes https://github.com/devfile/api/issues/334

**PR acceptance criteria**:

- [x] Test (WIP) 

- [x] Documentation (WIP)


.ci/run_tests_minikube_linux.sh
```diff
@@ -0,0 +1,40 @@
...
+# NOTE: This script assumes that minikube is installed and running, and using the docker driver on Linux
+# Due to networking issues with the docker driver and ingress on macOS/Windows, this script must be run on Linux
+
+# Share docker env with Minikube
+eval $(minikube docker-env)
+
+# error on unset variables
+set -u
+# print each command before executing it
+set -x
```
GeekArthur, line 12: Should we also add `set -e` to ensure the script can exit immediately if the command fails?

.ci/run_tests_minikube_linux.sh
```diff
@@ -0,0 +1,40 @@
...
+
+# Build the test devfile registry image
+./build_registry.sh
```
GeekArthur, line 15: Should we catch the error and exit here if the image build process is failed instead of continuing the deploy process?

GeekArthur (changes requested):

.ci/run_tests_minikube_linux.sh
johnmcollier: I left it out since it would prevent us from debugging via kubectl any failed deployments with helm.

.ci/run_tests_minikube_linux.sh
johnmcollier: Yes, that should be caught.

johnmcollier: Updated.

GeekArthur (approved):
LGTM, thanks for making the changes, it makes the dev process easier as well.

PR #64, opened 2021-06-23
Add basic troubleshooting doc for registry

johnmcollier:


TROUBLESHOOTING.md
```diff
@@ -0,0 +1,42 @@
...
+
+### Devfile Registry Does Not Start
+
+If the devfile registry fails to start, check the logs of both the devfile registry and oci registry containers. Any errors in the logs likely indicate that the devfile registry did not start properly.
+
+Potential causes of this include:
+
+- Specifying the wrong image for the devfile index image when deploying
+
+    - **Note:** The image produced by the devfile registry build tool must be used here if deploying your own devfile registry.
```
GeekArthur, line 29: `The image produced by devfile registry build tool` is a little vague I think, I think here you mean `quay.io/devfile/devfile-index-base:next`? It would be better we can let user know the specific image & tag that is produced by registry build tool.

TROUBLESHOOTING.md
```diff
@@ -0,0 +1,42 @@
+# Troubelshooting the Devfile Registry
+
+## Collecting Logs
+
+Logs can be collected for both the devfile index server, and the OCI registry server in the devfile registry, that should assist with debugging any potential issues.
+
+To retrieve the logs from the devfile index server:
+
+```bash
+kubectl logs <devfile-registry-pod> -c devfile-registry
```
GeekArthur, line 10: The container name is `devfile-registry-bootstrap` or `devfile-registry`? I am using operator to deploy the registry to my minikube cluster, the container name is `devfile-registry-bootstrap`.
```
kubectl logs devfile-registry-6675b84db4-gr5xf                                                                      1 ✘  minikube/test ⎈  14:18:40 
error: a container name must be specified for pod devfile-registry-6675b84db4-gr5xf, choose one of: [devfile-registry-bootstrap oci-registry]
```

johnmcollier: D'oh, you're right, I was looking at the hosted registry, which uses something different for _some_ reason.

TROUBLESHOOTING.md
johnmcollier: Not quite, `quay.io/devfile/devfile-index-base:next` is the base image.

I don't want to mention a specific image name or tag here, since it's going to depend on what the user pushes up to the registry. 

How about I just say "devfile index image"?

GeekArthur: Actually I rethink this, I think it's fine as from consumers' perspective they only have one image that is produced by the build tool and it's not necessary to specify that.

GeekArthur (approved):
LGTM

PR #73, opened 2021-07-29
Enable the devfile registry viewer

johnmcollier:
This PR enables the [devfile registry viewer](https://github.com/devfile/registry-viewer) in the devfile registry index server

- Updates the index server dockerfile to build the registry viewer
   - Base image changed to ubi8/nodejs-14-minimal
   - Since we're running the registry viewer under the `/viewer` endpoint, we have to build the viewer with `DEVFILE_VIEWER_ROOT=/viewer`
- Index server entrypoint updated to start the node server for the registry viewer, in addition to the index server
- Changes to index server endpoints:
   - `/`:
       - forwards to `/index` if no content type requested
       - forwards to`/viewer` if `text/html` is requested (i.e. in a browser)
   - `/viewer` and `/viewer/*` proxies all requests to`localhost:3000/viewer`
- Helm chart updates:
   - Config map to configure the devfile registry host for the viewer to `localhost:8080`
   - StartupProbe to verify `/viewer` is live
   - Slight tweaks to the readiness and liveness probe delays
   - Additional PRs will be needed in the https://github.com/devfile/registry-operator and https://github.com/devfile/registry repositories to update the operator and OpenShift templates with similar changes
- Integration tests in progress

Fixes https://github.com/devfile/api/issues/534

**PR acceptance criteria**:

- [ ] Test (WIP) 

- [ ] Documentation (WIP)


index/server/.dockerignore
```diff
@@ -0,0 +1,11 @@
+#registry-viewer dockerignore fields
+registry-viewer/tests
+registry-viewer/.vscode
+registry-viewer/.github
+registry-viewer/cypress
+registry-viewer/cypress.json
+registry-viewer/jest.config.js
+registry-viewer/jest.setup.js
+registry-viewer/LICENSE
+registry-viewer/OWNERSHIP
```
maysunfaisal, line 10: should be OWNERS, and no cypress

I wonder if there is a better way to handle this

johnmcollier: This is just copied from https://github.com/devfile/registry-viewer/blob/main/.dockerignore

TBH, I should probably just updated the build.sh script to copy this file over from registry-viewer, rather than maintaining a separate copy

index/server/Dockerfile
```diff
@@ -1,19 +1,34 @@
...
+RUN npm install -g yarn
+COPY registry-viewer . 
+
+ENV DEVFILE_VIEWER_ROOT /viewer
+ENV DEVFILE_COMMUNITY_HOST false
+
+# Build the devfile registry viewer
+RUN $(npm get prefix)/bin/yarn install --frozen-lockfile --ignore-optional
+RUN $(npm get prefix)/bin/yarn build
+RUN $(npm get prefix)/bin/yarn install --production --ignore-scripts --prefer-offline --ignore-optional
```
maysunfaisal, line 20: any idea how we can fetch the dockerfile steps from registry-viewer rather than having two copies of the same logic 🤔 (not that we need to do it now)

index/server/entrypoint.sh
```diff
@@ -10,7 +11,27 @@ if [ ! -e "$DEVFILE_INDEX" ]; then
...
 /registry/index-server
 
-fg %1
+
+# Try to force the registry viewer pages to generate
+# for n in {1..10}; do
+# 	echo "Sleeping $n"
+# 	sleep 1
+# 	curl http://localhost:8080/viewer || true
+# done
```
maysunfaisal, line 35: do we need all these uncommented info?

index/server/entrypoint.sh
```diff
@@ -10,7 +11,27 @@ if [ ! -e "$DEVFILE_INDEX" ]; then
...
+npm start &
+
+# Wait for server to start
+until $(curl --output /dev/null --silent --head --fail http://localhost:3000/viewer); do
+    printf '.'
+    sleep 1
+done
+
+# Copy over the registry viewer config
+# cp -rf /viewer-config/devfile-registry-hosts.json /app/config/
```
maysunfaisal, line 24: how do we set the local path using config in this case

index/server/entrypoint.sh
```diff
@@ -10,7 +11,27 @@ if [ ! -e "$DEVFILE_INDEX" ]; then
...
-fg %1
+
+# Try to force the registry viewer pages to generate
+# for n in {1..10}; do
+# 	echo "Sleeping $n"
+# 	sleep 1
+# 	curl http://localhost:8080/viewer || true
+# done
+
+tail -f /dev/null
```
maysunfaisal, line 37: just a thought, could the index server and the viewer pipe their output to a file and we can tail that file for logs?

index/server/entrypoint.sh
johnmcollier: This should be removed. The config is mounted directly to that folder as a volume via config map.

index/server/entrypoint.sh
johnmcollier: D’oh. I knew there was some leftover commented out stuff I needed to remove

index/server/entrypoint.sh
johnmcollier: So we don’t need to pipe the output since the output of the server is preserved with the`&`, so the output is still visible in the container logs.

But, this line is actually redundant, since we’re not running the index server in the background, so it can be removed.

index/server/Dockerfile
johnmcollier: You could do two separate docker builds:
1) Build the registry viewer image, using its dockerfile (build stage only, so need to specify that as a flag)
2) Build the index server. During the runtime stage, copy over the viewer resources from the viewer container built in step 1

There isn’t really a way to merge the two runtime stages (the index server and registry viewer) into one without doing something really messy, so I’d keep that separate

johnmcollier: I’ll see how feasible this is, if it’s nothing quick, I’ll shelve it for later

index/server/entrypoint.sh
johnmcollier: Fixed.

index/server/entrypoint.sh
johnmcollier: Fixed.

index/server/entrypoint.sh
johnmcollier: Fixed.

index/server/Dockerfile
johnmcollier: Fixed.

index/server/main.go
```diff
@@ -197,11 +198,35 @@ func main() {
 	router.HEAD("/v2/*proxyPath", ociServerProxy)
 	router.GET("/v2/*proxyPath", ociServerProxy)
 
+	router.Any("/viewer", serveUI)
+	router.Any("/viewer/*proxyPath", serveUI)
```
maysunfaisal, line 202: technically we only need GET for /viewer and all its sub paths right?

johnmcollier: Yeah. I used `Any` in case the viewer needed to support other methods in the future, but we can always change it back if needed.

I'll switch to GET, since we really only need that.

johnmcollier: Fixed.

johnmcollier:
PRs to update the openshift template and operator (to be merged after this pr is merged):
- OpenShift template: https://github.com/devfile/registry/pull/63
- Operator: https://github.com/devfile/registry-operator/pull/15

maysunfaisal (approved):
Screen Shot 2021-08-06 at 12 34 47 PM

working as expected

maysunfaisal (approved):

index/server/main.go
```diff
@@ -197,11 +198,35 @@ func main() {
...
+	router.GET("/viewer", serveUI)
+	router.GET("/viewer/*proxyPath", serveUI)
+
+	// Serve static content for stacks
 	router.Static("/stacks", stacksPath)
 
 	router.Run(":8080")
 }
 
+// serveDevfileIndex serves the index.json file located in the container at `serveDevfileIndex`
```
GeekArthur, line 210: We need to update the function comment for `serveRootEndpoint`

GeekArthur (changes requested):

index/server/build.sh
```diff
@@ -1,5 +1,16 @@
...
 
 # Build the index container for the registry
-buildfolder="$(basename "$(dirname "$0")")"
+buildfolder="$(realpath $(dirname $0))"
+
+# Clone the registry-support repo
+if [ -d $buildfolder/registry-viewer ]; then
+	rm -rf $buildfolder/registry-viewer
+fi
+git clone https://github.com/devfile/registry-viewer.git $buildfolder/registry-viewer
```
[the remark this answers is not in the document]
johnmcollier: I'd leave it as-is. Might make debugging a little more convenient.

index/server/entrypoint.sh
```diff
@@ -10,7 +11,14 @@ if [ ! -e "$DEVFILE_INDEX" ]; then
     exit 1
 fi
 
-# Start the index server
-/registry/index-server
+# Start the registry viewer
+npm start &
+
+# Wait for server to start
+until $(curl --output /dev/null --silent --head --fail http://localhost:3000/viewer); do
```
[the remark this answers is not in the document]
johnmcollier: Nah. The threshold already exists in the form of the startup/readiness/liveness probes which will fail if this goes on too long. Don't want to risk setting up a timeout that could fail before the probes themselves fail.

index/server/main.go
johnmcollier: Thanks. Updated.

GeekArthur (approved):

PR #74, opened 2021-08-03
Support retrieving sample devfiles from index server

johnmcollier:

This PR fixes the issue seen when requesting a sample's devfile from the `/devfiles/<id>` endpoint in the devfile registry index server where the server would crash. It also updates the endpoint to properly support retrieving sample devfiles now.

Fixes https://github.com/devfile/api/issues/549

**PR acceptance criteria**:

- [x] Test

### DIFF
--- a/index/server/main.go
+++ b/index/server/main.go
@@ -63,6 +63,7 @@ var mediaTypeMapping = map[string]string{
 
 var (
 	stacksPath            = os.Getenv("DEVFILE_STACKS")
+	samplesPath           = os.Getenv("DEVFILE_SAMPLES")
 	indexPath             = os.Getenv("DEVFILE_INDEX")
 	base64IndexPath       = os.Getenv("DEVFILE_BASE64_INDEX")
 	sampleIndexPath       = os.Getenv("DEVFILE_SAMPLE_INDEX")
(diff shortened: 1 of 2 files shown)


index/server/main.go
```diff
@@ -172,9 +173,18 @@ func main() {
...
+				} else {
+					// Retrieve the sample devfile stored under /registry/samples/<devfile>
+					sampleDevfilePath := path.Join(samplesPath, devfileIndex.Name, devfileName)
+					if _, err = os.Stat(sampleDevfilePath); err == nil {
+						bytes, err = ioutil.ReadFile(sampleDevfilePath)
+					}
+				}
+
 				if err != nil {
-					log.Fatal(err.Error())
```
GeekArthur, line 177: Should we directly remove it or use `log.Print(err.Error())` as errors will be saved in the server log?

johnmcollier: Yeah, we should still print it

johnmcollier: Fixed.

GeekArthur (approved):

maysunfaisal:
@johnmcollier FYI I think we should support cases where devfiles are not in the root dir https://github.com/maysunfaisal/node-bulletin-board-5/tree/main/app, its a current use case via the Console Import but can be done later on the registry-support

johnmcollier:
@maysunfaisal Can you get an issue open for that?

PR #75, opened 2021-08-17
Telemetry support

GeekArthur:
Signed-off-by: Jingfu Wang <jingfwan@redhat.com>

Registry index server

This PR implements the telemetry support so that we can monitor the metrics of consumer actions against the registry.

Related issue: https://github.com/devfile/api/issues/281

**PR acceptance criteria**:

- [x] Test (WIP) 

- [x] Documentation (WIP)

1. Deploy the registry
2. Access the registry REST APIs then check Segment and Woopra platform to monitor the registry metrics.


index/server/main.go
```diff
@@ -512,3 +604,50 @@ func buildIndexAPIResponse(c *gin.Context, indexType string, iconType string) {
...
+
+	err := client.Enqueue(event)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+// getUser gets the user
+func getUser(c *gin.Context) string {
```
johnmcollier, line 622: Can we use a different request header than user? Maybe client or something like that?

johnmcollier (changes requested):
Left some review comments.

index/server/main.go
johnmcollier: I also think we need to discuss how the registry clients will be (either via the go library or through rest) will be setting this header, and what the expected values will be.

johnmcollier:
Seeing the following integration test failures in the CI:
```
[Fail] [Verify oci registry is working properly] [It] /v2 endpoint should be available 
/registry-test/pkg/tests/ociregistry_tests.go:34

[Fail] [Verify oci registry is working properly] [It] /v2/_catalog endpoint should return a list of OCI artifacts 
/registry-test/pkg/tests/ociregistry_tests.go:40

[Fail] [Verify index server is working properly] [It] /devfiles/<devfile> endpoint should return a devfile for samples 
/registry-test/pkg/tests/indexserver_tests.go:119
```

The first two relate to this comment https://github.com/devfile/registry-support/pull/75#discussion_r690019422 and the last one is related to this comment https://github.com/devfile/registry-support/pull/75#discussion_r690017113

index/server/main.go
```diff
@@ -332,6 +404,27 @@ func ociServerProxy(c *gin.Context) {
 
 	// Set up the request to the proxy
 	// This is a good place to set up telemetry for requests to the OCI server (e.g. by parsing the path)
+	proxyPath := c.Param("proxyPath")
+	if proxyPath != "" {
+		name := strings.Split(proxyPath, "/")[2]
+		resource := strings.Split(proxyPath, "/")[3]
```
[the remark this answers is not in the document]
GeekArthur: I am not sure the version of the ORAS library that we use can try to access `/_catalog` endpoint and if `/_catalog` is valid, but I can add the validation code as part of string splitting process.

index/server/main.go
```diff
@@ -332,6 +404,27 @@ func ociServerProxy(c *gin.Context) {
...
+
+		// Track event for telemetry
+		if resource == "blobs" {
+			if enableTelemetry {
```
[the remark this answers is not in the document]
GeekArthur: You're right, will update it.

index/server/main.go
```diff
@@ -332,6 +404,27 @@ func ociServerProxy(c *gin.Context) {
...
+				err := trackEvent(analytics.Track{
+					Event:  eventTrackMap["download"],
+					UserId: "test",
```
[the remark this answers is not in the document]
GeekArthur: My bad, forget to remove the testing code.

index/server/main.go
```diff
@@ -463,3 +556,50 @@ func buildIndexAPIResponse(c *gin.Context, indexType string, iconType string) {
...
+func getUser(c *gin.Context) string {
+	user := defaultUser
+	if len(c.Request.Header["User"]) != 0 {
+		user = c.Request.Header["User"][0]
+	}
+	return user
+}
+
+// getOptionalEnv gets the optional environment variable
+func getOptionalEnv(key string, defaultValue interface{}) interface{} {
```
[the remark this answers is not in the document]
GeekArthur: Sure, I can try to add it.

index/server/main.go
GeekArthur: Sure, we can discuss that on part 2 which is client side change, this change is only for server side.

index/server/main.go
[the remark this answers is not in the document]
johnmcollier: ORAS may or may not necessarily access those endpoints, but those endpoints (among others) are still exposed, and we don't want the server to crash if they're hit

index/server/main.go
GeekArthur: The reason why we "user" here is to align with the terminology of Segment and Woopra, but discuss that later when I am working on the client side.

GeekArthur:
@johnmcollier Comments addressed, please review again, thanks.

index/server/main.go
```diff
@@ -345,12 +310,131 @@ func serveDevfileIndex(c *gin.Context) {
...
+				err := trackEvent(analytics.Track{
+					Event:  eventTrackMap["view"],
+					UserId: user,
+					Properties: analytics.NewProperties().
+						Set("name", name).
+						Set("type", string(devfileIndex.Type)).
+						Set("registry", registry),
+				})
+				if err != nil {
+					log.Print(err)
```
johnmcollier, line 425: Nit: `log.Println` instead.

index/server/main.go
```diff
@@ -345,12 +310,131 @@ func serveDevfileIndex(c *gin.Context) {
...
+
+		err := trackEvent(analytics.Track{
+			Event:  eventTrackMap["list"],
+			UserId: user,
+			Properties: analytics.NewProperties().
+				Set("type", indexType).
+				Set("registry", registry),
+		})
+		if err != nil {
+			log.Print(err)
```
johnmcollier, line 353: Nit: `log.Println` instead.

index/server/main.go
```diff
@@ -361,7 +445,38 @@ func ociServerProxy(c *gin.Context) {
...
 	// Set up the request to the proxy
-	// This is a good place to set up telemetry for requests to the OCI server (e.g. by parsing the path)
+	// Track event for telemetry
+	if enableTelemetry {
+		proxyPath := c.Param("proxyPath")
+		if proxyPath != "" {
+			var name string
+			var resource string
+			parts := strings.Split(proxyPath, "/")
+			if len(parts) == 5 {
```
johnmcollier, line 455: Why do we need to check for both lengths 4 and 5? Won't the proxyPath we want always be something like `/devfile-catalog/java-quarkus/blobs/sha256:d913cab108c3bc1bd06ce61f1e0cdb6eea2222a7884378f7e656fa26249990b9`, so the length we want would be 4? (`devfile-catalog`, `java-quarkus`, `blobs`, and `sha:...` in this example)

index/server/main.go
```diff
@@ -361,7 +445,38 @@ func ociServerProxy(c *gin.Context) {
...
+				name = parts[2]
+				resource = parts[3]
+			} else if len(parts) == 4 {
```
johnmcollier, line 458: It would be good to provide an example in the comments of the proxyPaths that we're parsing here to explain why we're checking for specfic lengths. E.g. just include an example like `/devfile-catalog/java-quarkus/blobs/sha256:d913cab108c3bc1bd06ce61f1e0cdb6eea2222a7884378f7e656fa26249990b9`

johnmcollier (changes requested):
Left a few more minor comments, nothing big though.

index/server/main.go
```diff
@@ -345,12 +310,131 @@ func serveDevfileIndex(c *gin.Context) {
 		timer.ObserveDuration()
 	}()
 
-	// Serve the index.json file
 	indexType := "stack"
 	iconType := c.Query("icon")
+
+	// Track event for telemetry
+	if enableTelemetry {
```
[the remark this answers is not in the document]
GeekArthur: Good point, will update

index/server/main.go
```diff
@@ -345,12 +310,131 @@ func serveDevfileIndex(c *gin.Context) {
...
 	buildIndexAPIResponse(c, indexType, iconType)
 }
 
+// serveDevfileIndexWithType returns the index file content with specific devfile type
+func serveDevfileIndexWithType(c *gin.Context) {
+	indexType := c.Param("type")
+	iconType := c.Query("icon")
+
+	// Track event for telemetry
+	if enableTelemetry {
```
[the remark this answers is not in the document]
GeekArthur: Good point, will update

index/server/main.go
GeekArthur: Currently it always be `/v2/devfile-catalog/nodejs/blobs/sha256:cc3bf36be550d38b701753da58034143da93ec387693f4ef98136e3e06e71fe1` (prepend API version), but I think you brought up we might receive `/devfile-catalog/*` (without API version) on previous code review, that's why I make the splitting part more robust to support the two APIs (check both lengths 4 and 5)

index/server/main.go
GeekArthur: Done

GeekArthur:
@johnmcollier I just address those comments, please review again.

index/server/main.go
johnmcollier: ProxyPath is the path that follows /v2, so I don’t think it will ever contain /v2 in it. So I just think we just need to check for an array of length 4

GeekArthur: You're right, the proxyPath doesn't contain the version, but the only length we need to check is 5 instead of 4. The reason is we use `split("/")` so the first element would be `""` which means for this line `parts := strings.Split(proxyPath, "/")`, if the proxyPath is `/devfile-catalog/java-quarkus/blobs/sha256:d913cab108c3bc1bd06ce61f1e0cdb6eea2222a7884378f7e656fa26249990b9` the parts array would be:

```
array[0]: ""
array[1]: devfile-catalog
array[2]: java-quarkus
array[3]: blobs
array[4]: sha256:d913cab108c3bc1bd06ce61f1e0cdb6eea2222a7884378f7e656fa26249990b9
```

I also test it by injecting some debug logs below
```
2021/08/23 15:44:45 proxyPath is /devfile-catalog/nodejs/blobs/sha256:cc3bf36be550d38b701753da58034143da93ec387693f4ef98136e3e06e71fe1
2021/08/23 15:44:45 parts len is 5
2021/08/23 15:44:45 name is nodejs, resource is blobs
```

with source code
```
                        parts := strings.Split(proxyPath, "/")
			log.Printf("parts len is %d\n", len(parts))
			// Check proxyPath with API version - "/v2/devfile-catalog/nodejs/blobs/sha256:cc3bf36be550d38b701753da58034143da93ec387693f4ef98136e3e06e71fe1"
			if len(parts) == 5 {
				name = parts[2]
				resource = parts[3]
				log.Printf("name is %s, resource is %s\n", name, resource)
			} else if len(parts) == 4 {
				// Check proxyPath without API version - "/devfile-catalog/java-quarkus/blobs/sha256:d913cab108c3bc1bd06ce61f1e0cdb6eea2222a7884378f7e656fa26249990b9"
				name = parts[1]
				resource = parts[2]
				log.Printf("name is %s, resource is %s\n", name, resource)
			}
```

GeekArthur: Updated

GeekArthur:
@johnmcollier Comments addressed, please review again, thanks

index/server/main.go
johnmcollier: 🤦‍♂️ You're right, I missed the empty space when I was testing.

johnmcollier (approved):

PR #80, opened 2021-09-02
Add dispatch payload

johnmcollier:
For more information, please see https://github.com/devfile/api/issues/527

This PR updates the dispatch workflow that we have to include the following in its payload:
- Repository name
- Commit SHA

This enables us to include the repository name and commit sha in the commit messages that triggers the repository deploy to staging.

It first checks if the the repository name and commit sha already exist in the workflow's client payload. If so, that means the workflow was triggered from the registry-viewer repo, and we want to use its repo name and commit sha instead. If it doesn't exist, that means a commit in this repo is triggering the deploy and we want to use it instead.

### DIFF
--- a/.github/workflows/pushimage-next.yaml
+++ b/.github/workflows/pushimage-next.yaml
@@ -45,6 +45,7 @@ jobs:
           token: ${{ secrets.REPO_ACCESS_TOKEN }}
           repository: ${{ matrix.repo }}
           event-type: build
+          client-payload: '{"repo": "${{ github.event.repository.name }}", "sha": "${{ github.sha }}"}'
 
   ociRegistryBuild:
     runs-on: ubuntu-latest
--- a/.github/workflows/pushimage-next.yaml
+++ b/.github/workflows/pushimage-next.yaml
@@ -42,12 +42,20 @@ jobs:
       - name: Get the repository name and commit sha
         id: get_commit_info
         run: |
+          # Check if repository name is not set in the client payload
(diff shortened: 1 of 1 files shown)


johnmcollier:
@GeekArthur Please review

johnmcollier:
An example workflow can be found here: https://github.com/johnmcollier/registry-support-test-workflow/runs/3496729903?check_suite_focus=true

With a resulting commit in the repo here: https://github.com/johnmcollier/registry-test-workflow/commits/main

.github/workflows/pushimage-next.yaml
```diff
@@ -39,12 +39,36 @@ jobs:
...
+        id: get_commit_info
+        run: |
+          # Check if repository name is not set in the client payload
+          # If it's not set, the triggered commit came from this repo (registry-support)
+          # If it is set, the triggered commit came from registry-viewer
+          if [ -z ${{ github.event.client_payload.repo }} ];
+          then
+            echo ::set-output name=repo::$(echo ${{ github.event.repository.name }})
+          else
+            echo ::set-output name=repo::$(echo ${{ github.event.client_payload.repo }})
```
maysunfaisal, line 52: I'm assuming `client_payload` is the github context syntax and this is the https://github.com/devfile/registry-viewer/blob/main/.github/workflows/push-changes.yaml#L20 dispatch action syntax?

johnmcollier: Yup.

maysunfaisal (approved):

PR #97, opened 2022-01-17
Access-Control-Allow-Origin header added to `/index*` routes

michael-valdron:

registry-support/server/index

Adds the `Access-Control-Allow-Origin: *` header to response of `/index/:type`. Without this will cause a CORS error in the case specified in the linked issue.

[devfile/api#719](https://github.com/devfile/api/issues/719)

**PR acceptance criteria**:

- [ ] Test (WIP) 

- [ ] Documentation (WIP)

### DIFF
--- a/index/server/pkg/server/endpoint.go
+++ b/index/server/pkg/server/endpoint.go
@@ -50,8 +50,6 @@ func serveDevfileIndex(c *gin.Context) {
 
 // serveDevfileIndexWithType returns the index file content with specific devfile type
 func serveDevfileIndexWithType(c *gin.Context) {
-	// Set headers
-	c.Header("Access-Control-Allow-Origin", "*")
 
 	// Serve the index with type
 	buildIndexAPIResponse(c)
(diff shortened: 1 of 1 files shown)


index/server/pkg/server/endpoint.go
```diff
@@ -50,6 +50,8 @@ func serveDevfileIndex(c *gin.Context) {
 
 // serveDevfileIndexWithType returns the index file content with specific devfile type
 func serveDevfileIndexWithType(c *gin.Context) {
+	// Set headers
+	c.Header("Access-Control-Allow-Origin", "*")
```
johnmcollier, line 54: Let's discuss on today's call if there's any other endpoints we'll want to add this for. Minimally I'm thinking the `/index` endpoints (so `serveDevfileIndex` as well).

johnmcollier (commented):
Generally looks good to me, but not approving yet due to the comment I left down below.

index/server/pkg/server/endpoint.go
michael-valdron: Sounds good.

johnmcollier (approved):

PR #104, opened 2022-02-28
Update telemetry enablement

kim-tsao:
Signed-off-by: Kim Tsao <ktsao@redhat.com>

Removes the hardcoded segment write key

We do not want to be collecting telemetry from users who set up their own devfile registries (per CNCF telemetry data policy).  They should be sending telemetry to their own Segment instance

Fixes #?
https://github.com/devfile/api/issues/769

**PR acceptance criteria**:

- [x] Test (WIP) 

- [ ] Documentation (WIP)

### DIFF
--- a/deploy/chart/devfile-registry/values.yaml
+++ b/deploy/chart/devfile-registry/values.yaml
@@ -32,5 +32,5 @@ persistence:
   size: 1Gi
 
 telemetry:
-  enabled: "false"
   registryName: "devfile-registry"
+  key: "6HBMiy5UxBtsbxXx7O4n0t0u4dt8IAR3"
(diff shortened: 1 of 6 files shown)


deploy/chart/devfile-registry/values.yaml
```diff
@@ -32,5 +32,5 @@ persistence:
   size: 1Gi
 
 telemetry:
-  enabled: "false"
   registryName: "devfile-registry"
+  key: "6HBMiy5UxBtsbxXx7O4n0t0u4dt8IAR3"
```
johnmcollier, line 36: Should this be unset instead?

johnmcollier (changes requested):
Generally looks good to me, can you also update the registry chart's readme to reference the new `values.telemetry.key` field?

deploy/chart/devfile-registry/values.yaml
kim-tsao: yes, let me update that

johnmcollier (approved):

PR #121, opened 2022-06-17
Skip offline test cases for registry acceptance tests

johnmcollier:
These test cases will have test data that won't be available when we run these as part of the acceptance tests in staging. So skipping them for now


michael-valdron (approved):
/lgtm

PR #122, opened 2022-06-20
Fix typo in build.sh

johnmcollier:
Fixes a typo in build.sh which prevents files from properly being tarred


michael-valdron (approved):
/lgtm

PR #128, opened 2022-07-14
Minor changes to the download cmd in handling the context path and bug fixes

michael-valdron:
Signed-off-by: Michael Valdron <mvaldron@redhat.com>

registry library

Recently added in PR #126, the `registry download` command downloads and extracts a starter project to a given context path (default `.`). At this time, it will create a sub-directory under the context path to extract the starter project contents to, however, it would be more useful and consistent with `registry pull` to extract the contents directly into the context path itself which is what this PR intents to provide. 

After this change, one can still create a sub-directory by specifying `--context=./<subdirectory_name>`. 

Other Changes:

- Bug fix to `DownloadStarterProjectAsDir` extract ZipSlip check, if the entire resultant file path is the just the file name (no directory prefix) do not perform check.
    - Without this fix the function will error in all cases of using the relative path to the present working directory `.`.
- Bug fix to `DownloadStarterProjectAsBytes` not downloading correctly if base registry url ends with `/`

Fixes #?

fixes devfile/api#892
fixes devfile/api#895

**PR acceptance criteria**:

- [ ] Test (WIP) 

Documentation (WIP)

### DIFF
--- a/registry-library/cmd/root.go
+++ b/registry-library/cmd/root.go
@@ -18,7 +18,6 @@ package cmd
 import (
 	"fmt"
 	"os"
-	"path/filepath"
 
 	"github.com/spf13/cobra"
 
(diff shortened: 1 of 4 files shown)


johnmcollier (commented):
Generally looks good. Do we need to add test cases for the bugfixes, or are the updates sufficient?

michael-valdron:
> Generally looks good. Do we need to add test cases for the bugfixes, or are the updates sufficient?

@johnmcollier Oh you mean test cases which say for example test that `https://registry.devfile.io` and `https://registry.devfile.io/` work the same or context = `.` does not error?

johnmcollier (approved):

PR #157, opened 2023-01-13
Add telemetry for the `download-starter-project` endpoint to registry server

michael-valdron:
Signed-off-by: Michael Valdron <mvaldron@redhat.com>

registry

Before, the download starter project process was only done in the registry viewer which tracked this telemetry metric itself https://github.com/devfile/registry-viewer/blob/145a485893b726808be40e1e31cc1ec96a8dde96/src/components/DevfilePage/DevfilePageProjects/DevfilePageProjects.tsx#L85.

In recent changes, the registry viewer no longer handles this process itself and instead uses an endpoint on the registry server (devfile/api#720). In addition, the registry viewer no longer tracks starter project downloads since devfile/api#747 was closed and therefore telemetry on starter project downloads needs to be tracked on the server endpoint itself. This PR provides these changes to resolve devfile/api#803.

Fixes #?

fixes devfile/api#803

**PR acceptance criteria**:

- [ ] Test Coverage 

Documentation (WIP)

### DIFF
--- a/index/server/pkg/server/endpoint.go
+++ b/index/server/pkg/server/endpoint.go
@@ -1,5 +1,5 @@
 //
-// Copyright 2022 Red Hat, Inc.
+// Copyright 2022-2023 Red Hat, Inc.
 //
 // Licensed under the Apache License, Version 2.0 (the "License");
 // you may not use this file except in compliance with the License.
--- a/index/server/pkg/server/endpoint.go
+++ b/index/server/pkg/server/endpoint.go
@@ -305,6 +305,26 @@ func serveDevfileStarterProjectWithVersion(c *gin.Context) {
 			return
 		}
 
+		// Track event for telemetry. Ignore events from the registry-viewer and DevConsole since those are tracked on the client side. Ignore indirect calls from clients.
+		if enableTelemetry && !util.IsWebClient(c) && !util.IsIndirectCall(c) {
+
+			user := util.GetUser(c)
+			client := util.GetClient(c)
+
+			err := util.TrackEvent(analytics.Track{
+				Event:   eventTrackMap["spdownload"],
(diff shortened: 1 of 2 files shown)


index/server/pkg/server/endpoint.go
```diff
@@ -305,6 +305,26 @@ func serveDevfileStarterProjectWithVersion(c *gin.Context) {
 			return
 		}
 
+		// Track event for telemetry. Ignore events from the registry-viewer and DevConsole since those are tracked on the client side. Ignore indirect calls from clients.
+		if enableTelemetry && !util.IsWebClient(c) && !util.IsIndirectCall(c) {
+
+			user := util.GetUser(c)
+			client := util.GetClient(c)
```
kim-tsao, line 312: are you passing in the client name `registry-viewer`?  We are ignoring those because we wanted to avoid tracking duplicate events .   Maybe you can add an exception?  Check if client is `registry-viewer` and `starterProjectName` is not empty?  Not sure if that's unique enough to infer it's a download event

kim-tsao (commented):
Also update the year range in license header for the modified files e.g. `2022-2023`

index/server/pkg/server/endpoint.go
michael-valdron: @kim-tsao As discussed, for the time being we can assume that the client is most likely the registry viewer if the client name is the generic `devfile-registry`. Other clients using this endpoint will be setting the `Client` header property to change this or using their own telemetry.

kim-tsao (commented):
@michael-valdron can you update the year in the license headers for these files?  Since they were created in `2022`, add the year range to indicate original and modified date e.g. `2022-2023`

kim-tsao (approved):

PR #173, opened 2023-04-18
Update directly executed scripts on workflows and scripts

thepetk:

/area ci

It ensures that all script executions, inside `.github/workflows` and inside scripts code, are not directly executed and that the script file is passed to the appropriate shell defined by the scripts shebang. For example. in the case of a bash script ./build.sh would become bash ./build.sh.

fixes https://github.com/devfile/api/issues/973

Fixes #?

**PR acceptance criteria**:

- [x] Test Coverage 

Documentation (WIP)

### DIFF
--- a/index/server/Dockerfile
+++ b/index/server/Dockerfile
@@ -38,4 +38,4 @@ USER 1001
 
 EXPOSE 8080
 
-ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT ["bash /entrypoint.sh"]
--- a/index/server/Dockerfile
+++ b/index/server/Dockerfile
@@ -38,4 +38,4 @@ USER 1001
 
 EXPOSE 8080
 
-ENTRYPOINT ["bash /entrypoint.sh"]
+ENTRYPOINT ["/entrypoint.sh"]
--- a/index/server/Dockerfile
+++ b/index/server/Dockerfile
@@ -11,6 +11,7 @@ USER root
 # Install and configure dependencies
 RUN microdnf update -y && microdnf install shadow-utils findutils && rm -rf /var/cache/yum
 COPY entrypoint.sh /
+RUN chmod +x /entrypoint.sh
 
 # Copy index server
 COPY --from=index-builder /tools/index-server /registry/index-server
--- a/.ci/Dockerfile
+++ b/.ci/Dockerfile
@@ -31,7 +31,7 @@ COPY tests/registry /registry
 RUN bash /build-tools/dl_starter_projects.sh go-starter community
 
 # Run the registry build tools
-RUN /build-tools/build.sh /registry /build
+RUN bash /build-tools/build.sh /registry /build
 
 FROM devfile-index-base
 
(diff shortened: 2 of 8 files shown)


michael-valdron (changes requested):
The following lines are missing `bash`:
- https://github.com/thepetk/devfile-registry-support/blob/21b3c798866953932547c0b60a035e26dde6f325/.ci/run_tests_minikube_linux.sh#L15
- https://github.com/thepetk/devfile-registry-support/blob/21b3c798866953932547c0b60a035e26dde6f325/.ci/run_tests_minikube_linux.sh#L43
- https://github.com/thepetk/devfile-registry-support/blob/21b3c798866953932547c0b60a035e26dde6f325/.ci/Dockerfile#L34

Also, either:
- Add `RUN chmod +x /entrypoint.sh` after https://github.com/thepetk/devfile-registry-support/blob/21b3c798866953932547c0b60a035e26dde6f325/index/server/Dockerfile#L13
**or**
- Add `bash` to https://github.com/thepetk/devfile-registry-support/blob/21b3c798866953932547c0b60a035e26dde6f325/index/server/Dockerfile#L41

index/server/Dockerfile
```diff
@@ -38,4 +38,4 @@ USER 1001
 
 EXPOSE 8080
 
-ENTRYPOINT ["/entrypoint.sh"]
+ENTRYPOINT ["bash /entrypoint.sh"]
```
michael-valdron, line 41: > Also, either:
> - Add `RUN chmod +x /entrypoint.sh` after https://github.com/thepetk/devfile-registry-support/blob/21b3c798866953932547c0b60a035e26dde6f325/index/server/Dockerfile#L13
> **or**
> - Add `bash` to https://github.com/thepetk/devfile-registry-support/blob/21b3c798866953932547c0b60a035e26dde6f325/index/server/Dockerfile#L41

My mistake, test is failing because of my last suggestion. Revert this and do the first one.
```suggestion
ENTRYPOINT ["/entrypoint.sh"]
```

michael-valdron (changes requested):

index/server/Dockerfile
thepetk: Yes I was investigating now :D Thanks for the help!

thepetk: Done :)

michael-valdron (changes requested):
> > Also, either:
> > 
> > * Add `RUN chmod +x /entrypoint.sh` after https://github.com/thepetk/devfile-registry-support/blob/21b3c798866953932547c0b60a035e26dde6f325/index/server/Dockerfile#L13
> >   **or**
> > * Add `bash` to https://github.com/thepetk/devfile-registry-support/blob/21b3c798866953932547c0b60a035e26dde6f325/index/server/Dockerfile#L41
> 
> My mistake, test is failing because of my last suggestion. Revert this and do the first one.

Could you add `RUN chmod +x /entrypoint.sh` after https://github.com/thepetk/devfile-registry-support/blob/2ba2ba41bad7daf4065b0e17418a88119f803a05/index/server/Dockerfile#L13. Still need this because of the assumption that the source `entrypoint.sh` will be non-executable.

michael-valdron (approved):
/lgtm

PR #178, opened 2023-08-03
Check for revisions upon Samples caching

thepetk:

/area registry

This PR is introducing a check for samples using a different branch than the default. More detailed we have two cases:

* A sample in `extraDevfileEntries.yaml`, having different versions while using the same repo. So the version X of a sample is using the X branch of sample repo, while the version Y of sample Y uses the branch Y of the same repo.

* A sample with a single version that is not using the default branch.

More detailed, a simple addition is made to the `cache_sample.sh` file in order to check the given `extraDevfileEntries.yaml` for any `git.checkoutFrom.revision` fields inside a sample. For this, we are introducing the `clone_sample_repo` function:

```bash
function clone_sample_repo() {
  local repoUrl="$1"
  local repoOutputDir="$2"
  local repoRevision="$3"
  git clone $repoUrl $repoOutputDir
  if [[ $repoRevision != "null" ]]; then
    cd $repoOutputDir && git checkout $repoRevision && cd -
  fi
}
```

As a result we can support samples defined like:
```yaml
name: go-basic
displayName: Basic Go
description: Go 1.16 application
icon: https://go.dev/blog/go-brand/Go-Logo/SVG/Go-Logo_Blue.svg
tags:
  - Go
projectType: Go
language: Go
provider: Red Hat
versions:
  - version: 1.0.2
    schemaVersion: 2.1.0
    git:
      checkoutFrom:
        revision: main
      remotes:
        origin: https://github.com/thepetk/devfile-sample-go-basic.git
  - version: 2.1.0
    schemaVersion: 2.2.0
    default: true
    git:
      checkoutFrom:
        revision: update_devfile
      remotes:
        origin: https://github.com/thepetk/devfile-sample-go-basic.git
```

Fixes https://github.com/devfile/api/issues/1209

**PR acceptance criteria**:

- [x] Test Coverage 

Documentation (WIP)

### DIFF
--- a/build-tools/cache_samples.sh
+++ b/build-tools/cache_samples.sh
@@ -5,11 +5,26 @@
 # The downloaded samples are cached under /registry/samples in the devfile registry container
 set -eu
 
-# download_sample takes in a given sample name (e.g. nodejs-basic), and git clones its corresponding repository
+# clone_sample_repo clones a given git repository to the sampleDir. Before cloning it checks
+# whether or not a revision exists.
+# Parameters:
+# 1: gitRepository  (git repository url)
+# 2: sampleDir      (output directory of the clone command)
+# 3: revision       (the prefered revision of the git repo we want to clone)
+function clone_sample_repo() {
+  local repoUrl="$1"
+  local repoOutputDir="$2"
+  local repoRevision="$3"
+  git clone $repoUrl $repoOutputDir
+  if [[ $repoRevision != "null" ]]; then
+    cd $repoOutputDir && git checkout $repoRevision && cd -
+  fi
+}
+
+# cache_sample takes in a given sample name (e.g. nodejs-basic), and git clones its corresponding repository
 # Parameters:
 # 1: Sample name (e.g. nodejs-basic)
-# 2: Path to extraDevfileEntries.yaml
-# 3: Output directory
+# 2: Output directory
 function cache_sample() {
     local sampleName="$1"
     local outputDir="$2"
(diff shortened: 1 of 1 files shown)


build-tools/cache_samples.sh
```diff
@@ -5,11 +5,27 @@
 # The downloaded samples are cached under /registry/samples in the devfile registry container
 set -eu
 
-# download_sample takes in a given sample name (e.g. nodejs-basic), and git clones its corresponding repository
+# clone_sample_repo clones a given git repository to the sampleDir. Before cloning it checks
+# wether or not a revision exists.
```
mike-hoang, line 9:
```suggestion
# whether or not a revision exists.
```

build-tools/cache_samples.sh
```diff
@@ -5,11 +5,27 @@
...
+# 2: sampleDir      (output directory of the clone command)
+# 3: revision       (the prefered revision of the git repo we want to clone)
+function clone_sample_repo() {
+  local repoUrl="$1"
+  local repoOutputDir="$2"
+  local repoRevision="$3"
+  if [[ $revision == "null" ]]; then
+    git clone $repoUrl $repoOutputDir
+  else
+    git clone --single-branch --branch $repoRevision $repoUrl $repoOutputDir
```
mike-hoang, line 21: This will only work for revisions where it's the branch name or tag and will fail for commit id.

```shell
➜  test git clone --single-branch --branch 605081dc84696eb31f217dec3c5be86a9ec38fc8 https://github.com/devfile/registry-support.git
Cloning into 'registry-support'...
warning: Could not find remote branch 605081dc84696eb31f217dec3c5be86a9ec38fc8 to clone.
fatal: Remote branch 605081dc84696eb31f217dec3c5be86a9ec38fc8 not found in upstream origin

➜  test git clone --single-branch --branch v1.0.1 https://github.com/devfile/registry-support.git
Cloning into 'registry-support'...
```

mike-hoang (commented):
We do allow commit ids to be used, so wondering if we should support that here?

ref: https://devfile.io/docs/2.2.0/devfile-schema#starter-projects-git-checkout-from-revision
```
revision string
The revision to checkout from. Should be branch name, tag or commit id. Default branch is used if missing or specified revision is not found.
```

build-tools/cache_samples.sh
thepetk: Oh yes nice catch! On it!

thepetk: I've updated the PR in order to checkout to a given revision instead of using `--single-branch` and `--branch` args

mike-hoang (approved):

PR #188, opened 2023-10-23
Update library version and fix revision usage in cache samples

thepetk:

/generator
/server

### Update devfile library version
This PR updates the version of `github.com/devfile/library/v2` to `v2.2.1`. One of the new features of library, will let us accept two different components using the same port. This way we can introduce new sample versions that will have consistent ports (using 8080 port).

With the current version of library the registry build fails as the validation of the devfile fails because of using twice the same port for different components. Trace:

```
* devfile contains multiple containers with same endpoint targetPort: 8080, imported from
id: go, registryURL: https://registry.devfile.io
```

### Use git.revision instead of git.checkoutFrom.revision
The PR also uses replaces the usage of `git.checkoutFrom.revision` inside the `cache_samples.sh` by using the `git.revision` which is the correct field according the `Git` schema: https://github.com/devfile/registry-support/blob/main/index/generator/schema/schema.go#L224

Partially fixes https://github.com/devfile/api/issues/1112

**PR acceptance criteria**:

- [x] Test Coverage 

Documentation (WIP)


tests/registry/stacks/go/2.1.0/devfile.yaml
```diff
@@ -11,4 +11,4 @@ metadata:
     - Go
   version: 2.1.0
 parent:
-  uri: https://registry.devfile.io/devfiles/go/2.1.0
+  uri: https://raw.githubusercontent.com/devfile/registry/main/stacks/go/2.1.0/devfile.yaml
```
maysunfaisal, line 14: is this change mean to be pushed?

maysunfaisal (approved):

tests/registry/stacks/go/2.1.0/devfile.yaml
thepetk: Yes, the test was failing due to the previous link. It seems that due to recent changes the URI of this format is not supported anymore. The reason is we are now using [DownloadGitRepoResources](https://github.com/devfile/library/blame/main/pkg/devfile/parser/utils.go#L44) which expects a git provider link instead of the registry one.

michael-valdron (approved):
/lgtm

PR #191, opened 2023-11-22
Disable http2 for indexServer

thepetk:

/server

This PR disables the HTTP2 protocol from the index server, following the instructions from here: https://pkg.go.dev/net/http#hdr-HTTP_2

This task is related to the mitigation of the Rapid Reset CVE. More info can be found in the related issue: https://github.com/devfile/api/issues/1342

Fixes https://github.com/devfile/api/issues/1342

**PR acceptance criteria**:

- [x] Test Coverage 

Documentation (WIP)


index/server/pkg/server/index.go
```diff
@@ -72,6 +73,7 @@ func ServeRegistry() {
 		Handler:      handler,
 		ReadTimeout:  10 * time.Second,
 		WriteTimeout: 10 * time.Second,
+		TLSNextProto: make(map[string]func(*http.Server, *tls.Conn, http.Handler)),
```
michael-valdron, line 76: Similar to what was done in https://github.com/devfile/registry-operator/pull/60, should we have an option for the registry administrator to enable http/2 and disable by default?

thepetk: Oh nice idea. I'll try to add the same flow so we can be consistent :) thanks @michael-valdron

thepetk: @michael-valdron I've added the `ENABLE_HTTP2` variable and tried to follow the same approach as the operator.

build_registry.sh
```diff
@@ -28,7 +28,7 @@ BASE_DIR=$(dirname $0)
 . ${BASE_DIR}/setenv.sh
 
 # Build the index server base image
-. ${BASE_DIR}/index/server/build.sh
+ENABLE_HTTP2="false" . ${BASE_DIR}/index/server/build.sh
```
michael-valdron, line 31: We should also accept optional http/2 enablement for the test registry build script as well.

thepetk:
As the current issue is related to security, I'd suggest to create a separate issue in order to improve the coverage on `index.go`

thepetk:
> As the current issue is related to security, I'd suggest to create a separate issue in order to improve the coverage on `index.go` (having in mind that the coverage diff is `~0,3%`)

I've created issue https://github.com/devfile/api/issues/1361 in order to keep track of the new test cases implementation

michael-valdron (approved):
/lgtm

PR #203, opened 2024-03-05
Fix devfile registry entry name and ingress url

michael-valdron:

Fixes the devfile registry entry name for registry viewer component of the helm chart. Fixed name 'Community' changed to value of `{{ template "devfileregistry.name" . }}` (the deployment name).

Additionally, the ingress url is now define and built under `_template.tpl`: https://github.com/michael-valdron/devfile-registry-support/blob/1df84bc08021587d954a3f6749699d77b40aaf28/deploy/chart/devfile-registry/templates/_template.tpl#L20-L27

Currently, this is only built from the ingress domain for k8s and devfile/api#1467 needs resolving to either separate this field or not rely on manually setting the domain for OpenShift environments.

Fixes #?

fixes devfile/api#1433

**PR acceptance criteria**:

- [X] Test Coverage 

Documentation (WIP)

### DIFF
--- a/deploy/chart/devfile-registry/templates/_template.tpl
+++ b/deploy/chart/devfile-registry/templates/_template.tpl
@@ -17,6 +17,15 @@
 {{- .Values.hostnameOverride | default (printf "devfile-registry-%s" .Release.Namespace) -}}
 {{- end -}}
 
+{{- define "devfileregistry.ingressUrl" -}}
+{{- $hostname := .Values.hostnameOverride | default (printf "devfile-registry-%s" .Release.Namespace) -}}
+{{- if .Values.global.tlsEnabled -}}
+{{- .Values.global.ingress.domain | printf "https://%s.%s" $hostname -}}
+{{- else -}}
+{{- .Values.global.ingress.domain | printf "http://%s.%s" $hostname -}}
+{{- end -}}
+{{- end -}}
+
 {{- define "devfileregistry.name" -}}
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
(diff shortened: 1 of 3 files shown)


michael-valdron:
> Documentation (WIP)
> 
> Does the [REST API doc](https://github.com/devfile/registry-support/index/server/registry-REST-API.adoc) need to be updated with your changes?
> Does the [registry library doc](https://github.com/devfile/registry-support/registry-library/README.md) need to be updated with your changes?

No changes needed.

deploy/chart/devfile-registry/templates/_template.tpl
```diff
@@ -17,6 +17,20 @@
 {{- .Values.hostnameOverride | default (printf "devfile-registry-%s" .Release.Namespace) -}}
 {{- end -}}
 
+{{- define "devfileregistry.ingressHostname" -}}
+{{- $hostname := .Values.hostnameOverride | default (printf "devfile-registry-%s" .Release.Namespace) -}}
+{{- .Values.global.ingress.domain | printf "%s.%s" $hostname -}}
```
Jdubrick, line 22: Will this format differ than the one from https://github.com/devfile/registry-operator/pull/80? On the operator side the hostname is `{hostName}.{ingressDomain}` but here it looks like `{ingressDomain}.{hostName}`. Is this the case and if so does it matter or should they be identical in formatting?

michael-valdron: @Jdubrick The piping `{{- .Values.global.ingress.domain | printf "%s.%s" $hostname -}}` passes `.Values.global.ingress.domain` into the last argument of `printf`, so its still `{hostName}.{ingressDomain}`. I guess I could do `{{- printf "%s.%s" $hostname .Values.global.ingress.domain -}}` instead since the piping seems unnecessary.

Jdubrick: Ah I was under the impression it was going first, in that case I think it's all good, the piping gets the job done I just misunderstood.

Jdubrick (approved):
/lgtm

PR #221, opened 2024-03-08
Helm Specs Update

Jdubrick:

This PR alters the timeout values so that the Helm deployment and operator deployment are synced in their values. Additionally it removes the env file mounts as the variables are defined in the spec and it is un-needed. This PR also updates the pod level security context to match that of the operator.

fixes https://github.com/devfile/api/issues/1430

**PR acceptance criteria**:

- [ ] Test Coverage 

Documentation (WIP)

Deployed Helm chart on both minikube and openshift and verified the security context was applied properly and deployment was successful.

### DIFF
--- a/deploy/chart/devfile-registry/templates/deployment.yaml
+++ b/deploy/chart/devfile-registry/templates/deployment.yaml
@@ -87,8 +87,8 @@ spec:
             port: 3000
             scheme: HTTP
           initialDelaySeconds: 30
-          periodSeconds: 1
-          timeoutSeconds: 10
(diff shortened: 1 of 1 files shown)


michael-valdron (approved):
/lgtm

Trying helm chart deployment works in my environment :heavy_check_mark: 

Noticed that the `viewer-env-file` was not removed from the `configmap,yaml` template, however this is also the case in the operator so this state is synced. Should open an issue to address the lingering configmap entry in both the helm chart and registry operator.

PR #232, opened 2024-04-10
Add deprecated stacks filtering to registry-library

thepetk:

/area registry-library

After the implementation for https://github.com/api/issues/959 we have now the ability of filtering the registry index for deprecated or non deprecated stacks by adding a `deprecated=true/false` parameter in the URL.

This PR updates the `registry-library` importing the same logic followed to the index server too.

Fixes https://github.com/devfile/api/issues/1065

**PR acceptance criteria**:

- [ ] Test Coverage 

Documentation (WIP)

### DIFF
--- a/registry-library/README.md
+++ b/registry-library/README.md
@@ -128,6 +128,22 @@ Supported devfile media types can be found in the latest version of [library.go]
       HTTPTimeout: &customTimeout
    }
    ```
+7. Filter *only deprecated* Devfiles based on the [lifecycle process for the devfiles project](https://github.com/devfile/registry/blob/main/LIFECYCLE.md)
+    ```go
+    options := registryLibrary.RegistryOptions{
+        Filter: registryLibrary.RegistryFilter{
+            Deprecated: "true",
+        },
+    }
+    ```
+8. Filter *only non deprecated* Devfiles based on the [lifecycle process for the devfiles project](https://github.com/devfile/registry/blob/main/LIFECYCLE.md)
+    ```go
+    options := registryLibrary.RegistryOptions{
+        Filter: registryLibrary.RegistryFilter{
+            Deprecated: "false",
+        },
+    }
+    ```
 
 #### Download the starter project
 
--- a/registry-library/README.md
+++ b/registry-library/README.md
@@ -132,7 +132,7 @@ Supported devfile media types can be found in the latest version of [library.go]
     ```go
     options := registryLibrary.RegistryOptions{
         Filter: registryLibrary.RegistryFilter{
-            Deprecated: "true",
+            Deprecated: registryLibrary.DeprecatedFilterTrue,
         },
     }
     ```
@@ -140,7 +140,7 @@ Supported devfile media types can be found in the latest version of [library.go]
     ```go
     options := registryLibrary.RegistryOptions{
         Filter: registryLibrary.RegistryFilter{
-            Deprecated: "false",
+            Deprecated: registryLibrary.DeprecatedFilterFalse,
         },
     }
     ```
--- a/registry-library/library/library.go
+++ b/registry-library/library/library.go
@@ -103,6 +103,10 @@ type RegistryFilter struct {
 	// only major version and minor version are required. e.g. 2.1, 2.2 ect. service version should not be provided.
 	// will only be applied if `NewIndexSchema=true`
 	MaxSchemaVersion string
+	// Deprecated is set to filter devfile index for stacks having the "Deprecated" tag inside their default set of tags
+	// or not. the only acceptable values are "true"/"false". in case the value is "true" it will only include only
+	// deprecated stacks, if is "false" only non deprecated. will only be applied if `NewIndexSchema=true`
+	Deprecated string
 }
 
 // GetRegistryIndex returns the list of index schema structured stacks and/or samples from a specified devfile registry.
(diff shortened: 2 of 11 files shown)


registry-library/library/library.go
```diff
@@ -103,6 +103,10 @@ type RegistryFilter struct {
 	// only major version and minor version are required. e.g. 2.1, 2.2 ect. service version should not be provided.
 	// will only be applied if `NewIndexSchema=true`
 	MaxSchemaVersion string
+	// Deprecated is set to filter devfile index for stacks having the "Deprecated" tag inside their default set of tags
+	// or not. the only acceptable values are "true"/"false". in case the value is "true" it will only include only
+	// deprecated stacks, if is "false" only non deprecated. will only be applied if `NewIndexSchema=true`
+	Deprecated string
```
michael-valdron, line 109: @thepetk I wondering if it would be good to use an "enum" for `RegistryFilter.Deprecated`, i.e. we create a `DeprecatedFilter` type that holds `string` values which `RegistryFilter.Deprecated` is typed as and create two constants:

```go
const DeprecatedFilterTrue DeprecatedFilter = "true"
const DeprecatedFilterFalse DeprecatedFilter = "false"
```

WDYT?

thepetk: I think makes sense! I've updated the code to reflect the suggested updates

registry-library/README.md
```diff
@@ -128,6 +128,22 @@ Supported devfile media types can be found in the latest version of [library.go]
...
+        Filter: registryLibrary.RegistryFilter{
+            Deprecated: "true",
+        },
+    }
+    ```
+8. Filter *only non deprecated* Devfiles based on the [lifecycle process for the devfiles project](https://github.com/devfile/registry/blob/main/LIFECYCLE.md)
+    ```go
+    options := registryLibrary.RegistryOptions{
+        Filter: registryLibrary.RegistryFilter{
+            Deprecated: "false",
```
michael-valdron, line 143:
```suggestion
            Deprecated: registryLibrary.DeprecatedFilterTrue,
        },
    }
    ```
8. Filter *only non deprecated* Devfiles based on the [lifecycle process for the devfiles project](https://github.com/devfile/registry/blob/main/LIFECYCLE.md)
    ```go
    options := registryLibrary.RegistryOptions{
        Filter: registryLibrary.RegistryFilter{
            Deprecated: registryLibrary.DeprecatedFilterFalse,
```

registry-library/library/library_test.go
```diff
@@ -315,6 +332,30 @@ func TestGetRegistryIndex(t *testing.T) {
...
+			devfileTypes: []indexSchema.DevfileType{indexSchema.StackDevfileType},
+			wantSchemas:  deprecatedFilteredIndex,
+		},
+		{
+			name: "Get Non Deprecated Filtered Index",
+			url:  "http://" + serverIP,
+			options: RegistryOptions{
+				NewIndexSchema: true,
+				Filter: RegistryFilter{
+					Deprecated: "false",
```
michael-valdron, line 353:
```suggestion
					Deprecated: DeprecatedFilterTrue,
				},
			},
			devfileTypes: []indexSchema.DevfileType{indexSchema.StackDevfileType},
			wantSchemas:  deprecatedFilteredIndex,
		},
		{
			name: "Get Non Deprecated Filtered Index",
			url:  "http://" + serverIP,
			options: RegistryOptions{
				NewIndexSchema: true,
				Filter: RegistryFilter{
					Deprecated: DeprecatedFilterFalse,
```

michael-valdron (changes requested):
I would reflect [these changes](https://github.com/devfile/registry-support/pull/232/commits/566efdf309edc3048cb504caf11eba18893d0ea1) in other areas as well (docs and testing):

michael-valdron (changes requested):

registry-library/library/library.go
```diff
@@ -103,6 +108,10 @@ type RegistryFilter struct {
 	// only major version and minor version are required. e.g. 2.1, 2.2 ect. service version should not be provided.
 	// will only be applied if `NewIndexSchema=true`
 	MaxSchemaVersion string
+	// Deprecated is set to filter devfile index for stacks having the "Deprecated" tag inside their default set of tags
+	// or not. the only acceptable values are "true"/"false". in case the value is "true" it will only include only
+	// deprecated stacks, if is "false" only non deprecated. will only be applied if `NewIndexSchema=true`
```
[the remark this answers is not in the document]
thepetk: I think this one is correct right? Only if we have the `NewIndexSchema=true` we will apply the filters.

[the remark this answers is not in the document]
michael-valdron: Yeah you are right :+1:

michael-valdron (approved):
/lgtm

thepetk:
> Change the value of [index/server/tests/registry/stacks/java-springboot/devfile.yaml#L9](https://github.com/devfile/registry-support/blob/b950a067ef2eb2a1bb08ea1658bedeb1b600e722/index/server/tests/registry/stacks/java-springboot/devfile.yaml#L9) to our sprintboot logo to get around this `https://raw.githubusercontent.com/devfile-samples/devfile-stack-icons/main/spring.svg`

I've added this change. Although I think better if we open an issue for a more permanent solution around this. I think a good approach here could be to try to gather all test resources into a single directory, shared between the modules of the repo. WDYT?

michael-valdron (approved):
/lgtm

PR #237, opened 2024-04-29
Feature: Deployment scopes field and filter

michael-valdron:

Adds the `deploymentScopes` field to the index schema and version schema with validation if field is set. Additionally, includes field filter to use field filtering implemented in devfile/api#959 to fuzzy filter based on detected deployment scope.

These changes **does not include** the detection of the deployments scopes, which are to be done in devfile/api#1197.

Additional Changes:
- [Fixes naming convention of filter parameter name constants](https://github.com/devfile/registry-support/pull/237/commits/56d075efe47d1b6c1f55ebabe448bf308004a4c3)

Fixes #?

fixes devfile/api#1321

**PR acceptance criteria**:

- [x] Test Coverage 

Documentation (WIP)


index/generator/library/library.go
```diff
@@ -193,6 +222,25 @@ func validateIndexComponent(indexComponent schema.Schema, componentType schema.D
...
+	if len(indexComponent.DeploymentScopes) > 2 {
+		return &TooManyDeploymentScopes{devfile: indexComponent.Name}
+	}
+	for kind := range indexComponent.DeploymentScopes {
+		if kind != schema.InnerloopKind && kind != schema.OuterloopKind {
+			return &InvalidDeploymentScopes{devfile: indexComponent.Name, deploymentScopeKind: kind}
+		}
+	}
+	for _, version := range indexComponent.Versions {
+		if len(version.DeploymentScopes) > 2 {
```
thepetk, line 234: @michael-valdron I was wondering what would be the case that a version will have 2 deployment scopes. I'm thinking this could be innerloop or outerloop right?

thepetk: I guess is based on the same format like CommandGroups right?

thepetk (commented):
Left a minor comment just to confirm a minor detail. In general it looks good to me (tests are passing, I also have build the new base image without a problem).

On a side note, I think the renaming of the parameters is out of scope. As far as I have tested I think they look good to me.

index/generator/library/library.go
michael-valdron: > @michael-valdron I was wondering what would be the case that a version will have 2 deployment scopes. I'm thinking this could be innerloop or outerloop right?

@thepetk Correct, every version (and stack) can only have 2 deployment scopes max, innerloop and outerloop.

michael-valdron:
> On a side note, I think the renaming of the parameters is out of scope.

Yeah, this was an extra change I did as I noticed the naming convention was not consistent with the rest of the constants.

thepetk (approved):
/lgtm

PR #238, opened 2024-05-07
Helm Chart OpenShift Route Handling

Jdubrick:
Bug fix and small enhancement

This PR aims to address the issue that occurs when a user installs the `devfile registry` via Helm to an OpenShift environment. Prior to this change the `fqdn` of the `registry-viewer` was not properly being set as it was expecting `global.ingress.domain` to be set, but if you are installing to OpenShift it was not a requirement to set that.

In order to solve this problem the following was added:
1. A helper script for OpenShift installations via Helm
   - This script allows a user to have the `fqdn` properly set if OpenShift auto generates a domain for you, or if you set a domain yourself. It utilizes `Helm Upgrade` to mimic the functionality currently in place for the `registry operator` where the `fqdn` is set after deployment (it spins up a new replacement pod after the information is known).

2. Template updates
   - In order to have more clarity and consistency, the template used by Helm now has portions for both `OpenShift Routes` and `Kubernetes Ingresses`.

3. Updated documentation
   - As a result of changes, documentation was updated to accurately reflect the different ways you can install the `devfile registry` with the use of a Helm chart.

Fixes https://github.com/devfile/api/issues/1467

**PR acceptance criteria**:

- [ ] Test Coverage 

Documentation (WIP)

The following cases will result in the `fqdn` of the `registry-viewer` being properly set:
1. Install to OpenShift with the use of the new script WITH & WITHOUT setting a domain
2. Install to OpenShift using the Helm CLI WITHOUT setting a domain
3. Install to Kubernetes using the Helm CLI WITH & WITHOUT setting a domain

The following case will result in the `fqdn` not being properly set:
1. Install to OpenShift with the Helm CLI WITH setting a domain

### DIFF
--- a/deploy/chart/devfile-registry/README.md
+++ b/deploy/chart/devfile-registry/README.md
@@ -21,17 +21,84 @@ E.g. if your cluster's ingress domain is 192.168.1.0.nip.io, you would run:
 helm install devfile-registry deploy/chart/devfile-registry --set global.ingress.domain=192.168.1.0.nip.io
 ```
 
+### Kubernetes Installation Examples
+```bash
+$ helm install devfile-registry ./deploy/chart/devfile-registry \ 
+    --set global.ingress.domain=<ingress-domain> \
+	--set devfileIndex.image=<index-image> \
+	--set devfileIndex.tag=<index-image-tag>
+```
+
+Where `<ingress-domain>` is the ingress domain for your cluster, `<index-image>` is the devfile index image you want to deploy, and `<index-image-tag>` is the corresponding image tag for the devfile index image.
+
+For example, if you're installing your own custom devfile registry image for dev/test purposes on Minikube, you might run:
+
+```bash
+$ helm install devfile-registry ./deploy/chart/devfile-registry \
+    --set global.ingress.domain="$(minikube ip).nip.io" \
+	--set devfileIndex.image=quay.io/someuser/devfile-index \
+	--set devfileIndex.tag=latest
+```
+
+You can deploy a devfile registry with a custom registry viewer image (uses `quay.io/devfile/registry-viewer:next` by default) by running the following:
+
+```bash
+$ helm install devfile-registry ./deploy/chart/devfile-registry \
+    --set global.ingress.domain="$(minikube ip).nip.io" \
+	--set devfileIndex.image=quay.io/someuser/devfile-index \
+	--set devfileIndex.tag=latest \
+	--set registryViewer.image=quay.io/someuser/registry-viewer \
+	--set registryViewer.tag=latest
+```
+
+You can deploy a *headless* devfile registry (i.e. without the registry viewer) by specifying `--set global.headless=true` which will look like:
+
+```bash
+$ helm install devfile-registry ./deploy/chart/devfile-registry \
+    --set global.ingress.domain="$(minikube ip).nip.io" \
+	--set global.headless=true \
+	--set devfileIndex.image=quay.io/someuser/devfile-index \
+	--set devfileIndex.tag=latest
+```
+
 ## Installing the Devfile Registry on OpenShift
 
-If you're installing on OpenShift, you need to set `global.isOpenShift` to true, for example:
+If you're installing on OpenShift, you will first need to set `global.isOpenShift` to true, for example:
 ```
 helm install devfile-registry deploy/chart/devfile-registry --set global.isOpenShift=true
 ```
 
-or, if you want to install a specific devfile index image, you can run:
+There are 3 ways that you can install the registry on OpenShift:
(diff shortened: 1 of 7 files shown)


deploy/chart/devfile-registry/README.md
```diff
@@ -21,17 +21,84 @@ E.g. if your cluster's ingress domain is 192.168.1.0.nip.io, you would run:
...
+    --set global.ingress.domain="$(minikube ip).nip.io" \
+	--set global.headless=true \
+	--set devfileIndex.image=quay.io/someuser/devfile-index \
+	--set devfileIndex.tag=latest
+```
+
 ## Installing the Devfile Registry on OpenShift
 
-If you're installing on OpenShift, you need to set `global.isOpenShift` to true, for example:
+If you're installing on OpenShift, you will first need to set `global.isOpenShift` to true, for example:
```
johnmcollier, line 66: Nit: We're setting `isOpenShift` to true during install, so I don't think it's been set _first_?

deploy/chart/devfile-registry/templates/_template.tpl
```diff
@@ -22,6 +22,15 @@
 {{- .Values.global.ingress.domain | printf "%s.%s" $hostname -}}
 {{- end -}}
 
+{{- define "devfileregistry.routeHostname" -}}
+{{- $hostname := .Values.hostnameOverride | default (printf "devfile-registry-%s" .Release.Namespace) -}}
```
johnmcollier, line 26: The route's hostname will be prefixed with `{{ template "devfileregistry.fullname" . }}`, which should default to `devfile-registry` for most cases. But to cover all grounds, should we change the default here to `{{ template "devfileregistry.fullname" . }}`?

Edit: I didn't see that the script below hardcoded the chart name to `devfile-registry`.

deploy/chart/devfile-registry/README.md
```diff
@@ -21,17 +21,84 @@ E.g. if your cluster's ingress domain is 192.168.1.0.nip.io, you would run:
...
 ```
 helm install devfile-registry deploy/chart/devfile-registry --set global.isOpenShift=true
 ```
 
-or, if you want to install a specific devfile index image, you can run:
+There are 3 ways that you can install the registry on OpenShift:
+#### 1: Via Installation Script With OpenShift Generated Route Hostname and Domain
```
johnmcollier, line 72: We don't necessarily need to cover this now, but something to keep in mind If we eventually onboard the chart to something like ArtifactHub, we would need to provide instructions for installing the chart _without_ the script. (So basically installing the chart, then covering how to get the route, and run `helm upgrade`).

Given that we don't run on artifacthub yet, I don't think we need to worry about that _just_ yet, but good to be aware of it.

helm-openshift-install.sh
```diff
@@ -0,0 +1,49 @@
...
+    echo "Usage: $0 <arguments>"
+    exit 1
+fi
+
+START_TIME=$(date +%s)
+TIMEOUT_SEC=25 
+SPINUP_SEC=60 # Helm Upgrade can have unexpected results if it is run while the cluster is initially starting
+
+# Run the install without the generated Openshift Route being passed to Registry Viewer fqdn as it does not exist yet
+helm install devfile-registry deploy/chart/devfile-registry --set global.isOpenShift=true "$@"
```
johnmcollier, line 28: Should `devfile-registry` be customizable here?

johnmcollier (commented):
Just a couple small comments

deploy/chart/devfile-registry/templates/_template.tpl
Jdubrick: With your edit do you still want to change the default?

deploy/chart/devfile-registry/README.md
Jdubrick: I think opening an issue for this to make sure it doesn't get lost would be good, wdyt?

helm-openshift-install.sh
Jdubrick: I can definitely edit it to allow the `devfile-registry` or name in this case be set via an env variable or something, same with the path if you think allowing that to change is also necessary. I suggest an environment variable instead of an argument to the script so we can continue using the simple passing in of the helm install arguments via `"$@". If we allow the name and path to be set via arguments I believe that would be fine we'd just need to parse the arguments first to determine where each needs to go

johnmcollier: Perhaps we can track it separately? I think it's probably fine to leave it.

deploy/chart/devfile-registry/templates/_template.tpl
johnmcollier: If we do the change, yes, but I think we can handle it separately.

johnmcollier (approved):

PR #239, opened 2024-05-14
Allow samples cache directory to be created if non-existent

michael-valdron:

build-tools

Changes `mkdir $samplesDir/$sample` to `mkdir -p $samplesDir/$sample` to allow target cache directories which do not exist yet.

Fixes #?

fixes devfile/api#1553

**PR acceptance criteria**:

- [ ] Test Coverage 

Documentation (WIP)

Run `bash build-tools/cache_samples.sh extraDevfileEntries.yaml <non_existent_caching_path>`

### DIFF
--- a/build-tools/cache_samples.sh
+++ b/build-tools/cache_samples.sh
@@ -99,7 +99,7 @@ function cache_devfile() {
     fi
 }
 
-if [[ "$OSTYPE" == "darwin" ]]
+if [[ "$OSTYPE" == "darwin"* ]]
 then
   if [[ $1 != ./* ]] && [[ $1 != ../* ]] && [[ $1 != /* ]]
   then
--- a/build-tools/cache_samples.sh
+++ b/build-tools/cache_samples.sh
@@ -103,7 +103,7 @@ devfileEntriesFile=$1
 samplesDir=$2
 
 for sample in $(yq e '(.samples[].name)' $devfileEntriesFile); do
-  mkdir $samplesDir/$sample
+  mkdir -p $samplesDir/$sample
   echo $sample
   cache_sample $sample $samplesDir/$sample
 done
--- a/build-tools/cache_samples.sh
+++ b/build-tools/cache_samples.sh
@@ -99,8 +99,24 @@ function cache_devfile() {
     fi
 }
 
-devfileEntriesFile=$1
-samplesDir=$2
+if [[ "$OSTYPE" == "darwin" ]]
+then
+  if [[ $1 != ./* ]] && [[ $1 != ../* ]] && [[ $1 != /* ]]
+  then
+    devfileEntriesFile=./$1
+  else
+    devfileEntriesFile=$1
+  fi
+  if [[ $2 != ./* ]] && [[ $2 != ../* ]] && [[ $2 != /* ]]
+  then
+    samplesDir=./$2
+  else
+    samplesDir=$2
+  fi
+else
+  devfileEntriesFile=$1
+  samplesDir=$2
+fi
 
 for sample in $(yq e '(.samples[].name)' $devfileEntriesFile); do
   mkdir -p $samplesDir/$sample


michael-valdron:
> - [ ] Test Coverage 
>     - Are your changes sufficiently tested, and are any applicable test cases added or updated to cover your changes?
> 
> Documentation (WIP)
> - [ ] Does the [REST API doc](../index/server/registry-REST-API.adoc) need to be updated with your changes?
> - [ ] Does the [registry library doc](../registry-library/README.md) need to be updated with your changes?

N/A

build-tools/cache_samples.sh
```diff
@@ -103,7 +103,7 @@ devfileEntriesFile=$1
 samplesDir=$2
 
 for sample in $(yq e '(.samples[].name)' $devfileEntriesFile); do
-  mkdir $samplesDir/$sample
+  mkdir -p $samplesDir/$sample
```
Jdubrick, line 106:
```suggestion
  mkdir -p .$samplesDir/$sample
```

When trying to test it locally I was running into the following error:
`mkdir: /samples: Read-only file system`

After doing some digging it seems it may be a MacOS issue, adding the `.` to tell the script it is creating the directories based on the location of the script fixed the issue, I think it is something to do with MacOS and roots. When you exclude the `.` it thinks you are creating from the absolute root, see [this StackOverflow post](https://stackoverflow.com/a/62818209) about it.

Not sure if that messes with the functionality of the script for other OS though.

michael-valdron: Addressing this here: https://github.com/devfile/registry-support/pull/239/commits/5bb20c06266bb35d446f2370fd53de400d557921

build-tools/cache_samples.sh
```diff
@@ -99,11 +99,27 @@ function cache_devfile() {
     fi
 }
 
-devfileEntriesFile=$1
-samplesDir=$2
+if [[ "$OSTYPE" == "darwin" ]]
```
Jdubrick, line 102:
```suggestion
if [[ "$OSTYPE" == "darwin"* ]]
```

I noticed it wasn't hitting the condition properly and I found if I run `echo $OSTYPE` I was receiving `darwin23.0`, adding the `*` allows for it to parse darwin properly if version is attached to it. Another option is to use `"$(uname -s)" == "Darwin"`

build-tools/cache_samples.sh
```diff
@@ -99,11 +99,27 @@ function cache_devfile() {
...
+then
+  if [[ $1 != ./* ]] && [[ $1 != ../* ]] && [[ $1 != /* ]]
+  then
+    devfileEntriesFile=./$1
+  else
+    devfileEntriesFile=$1
+  fi
+  if [[ $2 != ./* ]] && [[ $2 != ../* ]] && [[ $2 != /* ]]
```
Jdubrick, line 110:
```suggestion
  if [[ $2 != ./* ]] && [[ $2 != ../* ]]
```

If my `$samplesdir` was `/samples/.cache` for example it wasn't adding the `.` to make it `./samples/.cache` in this case

Jdubrick (changes requested):
After these changes I was able to run it on MacOS

build-tools/cache_samples.sh
michael-valdron: > If my `$samplesdir` was `/samples/.cache` for example it wasn't adding the `.` to make it `./samples/.cache` in this case

@Jdubrick I think this should be the case though, if one explicitly specifies `/samples/.cache` it shouldn't assume you mean a relative path. Correct me if I'm misunderstanding.

Jdubrick: @michael-valdron my from understanding when I pass `/samples/.cache` as the argument from the working directory when the script tries to create the directory it gets blocked by the OS on Mac because it assumes you are creating the directory from root. It may just be a Mac limitation that it has to assume you mean relative to the working directory? wdyt?

Jdubrick: Actually, I guess it could just be passed as `./samples/.cache` in this case to combat that, I believe you're right

Jdubrick: > Actually, I guess it could just be passed as `./samples/.cache` in this case to combat that, I believe you're right

Can confirm this is the case, I think this is good in that case!

Jdubrick (approved):
/lgtm

build-tools/cache_samples.sh
michael-valdron: resolved with https://github.com/devfile/registry-support/pull/239/commits/c1320acb37e95268aa78d0f8c4cd74669b1b52f6

PR #240, opened 2024-05-16
Multi-Arch Support for oci-registry builds and CI

Jdubrick:
Updating CI to support multi-arch image building

This PR contains the following changes:
- Updated `build.sh` script to allow for easy building of different architecture images, by default `linux/amd64` will be built with no arguments passed
- Updated `oci-registry/README.md` to reflect the additional functionality of `build.sh`
- Updated `ci.yaml` to include QEMU for Docker so that the workflow can build `linux/arm64` images
- Updated `ci.yaml` to test the builds for both `linux/amd64` and `linux/arm64` `oci-registry` images
- Created `build-multi-arch.sh` to enable multi-arch building and pushing for the `oci-registry` images. This by default will use Docker as github workflows supports that, however, it also allows for podman to be used if we ever switch. This script handles the building of both `linux/amd64` and `linux/arm64` images, the creation of the image manifest, and the pushing of all 3 items to `quay.io/devfile/oci-registry`. The architecture specific images will be tagged `amd64` and `arm64` for their respective images, and the manifest will be tagged `next`. Users should see no change in their workflow for pulling the `quay.io/devfile/oci-registry:next` image, it will pull the proper architecture for the OS pulling.
- Updated `pushimage-next.yaml` to include QEMU for Docker as well as run the multi-arch script.

fixes https://github.com/devfile/api/issues/1546

**PR acceptance criteria**:

- [ ] Test Coverage 

Documentation (WIP)

### DIFF
--- a/.github/workflows/pushimage-next.yaml
+++ b/.github/workflows/pushimage-next.yaml
@@ -89,6 +89,8 @@ jobs:
     steps:
       - name: Check out registry support source code
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
+      - name: Set up QEMU # Enables arm64 image building
+        uses: docker/setup-qemu-action@68827325e0b33c7199eb31dd4e31fbe9023e06e3 #v3.0.0
       - name: Login to Quay.io
         uses: docker/login-action@v3
         with:
--- a/.github/workflows/pushimage-next.yaml
+++ b/.github/workflows/pushimage-next.yaml
@@ -17,78 +17,20 @@ name: Next Dockerimage
 
 on:
   push:
-    branches: [main]
+    branches: [main, feat/multi-arch-support]
   repository_dispatch:
     types: [build]
 
 jobs:
-  indexServerBuild:
-    runs-on: ubuntu-latest
-    steps:
-      - name: Check out registry support source code
-        uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
-      - name: Setup Go environment
-        uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4.1.0
-        with:
-          go-version: 1.19
-      - name: Login to Quay
-        uses: docker/login-action@465a07811f14bebb1938fbed4728c6a1ff8901fc # v2.2.0
-        with:
-          registry: quay.io
-          username: ${{ secrets.QUAY_USERNAME }}
-          password: ${{ secrets.QUAY_PASSWORD }}
-      - name: Build the index server base image
-        run: | 
-          cd index/server
-          export GOPATH=$(go env GOPATH)
-          go install github.com/deepmap/oapi-codegen/cmd/oapi-codegen@v1.12.4
-          bash ./build.sh
-      - name: Push the index server base image
-        run: cd index/server && bash ./push.sh quay.io/devfile/devfile-index-base:next
-
-  dispatch:
-    needs: indexServerBuild
-    strategy:
-      matrix:
-        repo: ['devfile/registry']
-    runs-on: ubuntu-latest
-    steps:
-      - name: Get the repository name and commit sha
-        id: get_commit_info
-        run: |
-          # Check if repository name is not set in the client payload
-          # If it's not set, the triggered commit came from this repo (registry-support)
-          # If it is set, the triggered commit came from registry-viewer
-          if [ -z ${{ github.event.client_payload.repo }} ];
-          then
-            echo ::set-output name=repo::$(echo ${{ github.event.repository.name }})
-          else
-            echo ::set-output name=repo::$(echo ${{ github.event.client_payload.repo }})
-          fi
-
-          # Check if commit sha is not set in the client payload
-          # If it's not set, the triggered commit came from this repo (registry-support)
-          # If it is set, the triggered commit came from registry-viewer
-
-          if [ -z ${{ github.event.client_payload.sha }} ];
-          then
-            echo ::set-output name=sha::$(echo ${{ github.sha }})
-          else
-            echo ::set-output name=sha::$(echo ${{ github.event.client_payload.sha }})
-          fi
-      - name: Repository dispatch
-        uses: peter-evans/repository-dispatch@v1
-        with:
-          token: ${{ secrets.REPO_ACCESS_TOKEN }}
-          repository: ${{ matrix.repo }}
-          event-type: build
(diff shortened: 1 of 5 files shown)


.github/workflows/pushimage-next.yaml
```diff
@@ -89,17 +89,16 @@
     steps:
       - name: Check out registry support source code
         uses: actions/checkout@c85c95e3d7251135ab7dc9ce3241c5835cc595a9 # v3.5.3
-      - name: Build and push oci-registry docker image
-        uses: docker/build-push-action@v1.1.0
+      - name: Set up QEMU # Enables arm64 image building
+        uses: docker/setup-qemu-action@68827325e0b33c7199eb31dd4e31fbe9023e06e3 #v3.0.0
+      - name: Login to Quay.io
+        uses: docker/login-action@v3
```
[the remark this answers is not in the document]
Jdubrick: https://github.com/devfile/registry-support/pull/240/commits/8290e5b3a390d23a81414e477673dad86c464dbc

oci-registry/build-multi-arch.sh
```diff
@@ -0,0 +1,80 @@
...
+  engine-handler podman
+
+  # Create manifest and add images
+  podman manifest create oci-registry-manifest
+  for img in "${images[@]}" ; do
+    podman manifest add oci-registry-manifest "$img"
+  done
+
+  # Push and delete local manifest
+  podman manifest push oci-registry-manifest "$BASE_REPO":"$BASE_TAG"
```
thepetk, line 64: QQ: We push the image having a different tag. Is this expected?

thepetk: Do you think it might be better to assign the "$BASE_REPO:$BASE_TAG" to a variable so we can use the variable instead?

oci-registry/build-multi-arch.sh
```diff
@@ -0,0 +1,80 @@
...
+  podman manifest rm oci-registry-manifest
+
+else
+  echo "Executing with docker"
+
+  # Build and push multi-arch images
+  engine-handler docker
```
thepetk, line 71: I was testing locally the script (replaced with my quay.io workspace) and I'm getting
```
Dockerfile:18
--------------------
  16 |     
  17 |     FROM registry.access.redhat.com/ubi8-minimal:8.2
  18 | >>> RUN microdnf update -y && rm -rf /var/cache/yum && microdnf install ca-certificates httpd-tools
  19 |     
  20 |     # Create a non-root user to run the server as
--------------------
ERROR: failed to solve: process "/bin/sh -c microdnf update -y && rm -rf /var/cache/yum && microdnf install ca-certificates httpd-tools" did not complete successfully: exit code: 1
```

thepetk: Not sure tho if is something I'm missing. Tried:

* bash oci-registry/build-multi-arch.sh amd64
* bash oci-registry/build-multi-arch.sh

thepetk (commented):
Tested locally the scripts and left some comments. One question in general I have is for the manifests and their importance on the script. Mostly because while I've seen this feature in the pas is the first time I see it in action :D

oci-registry/build-multi-arch.sh
Jdubrick: > QQ: We push the image having a different tag. Is this expected?

For multi-arch when we build for `linux/arm64` and `linux/amd64` for example we will have 3 items at the end of the "build" stage:
1. An arm64 image for `oci-registry`
2. an amd64 image for `oci-registry`
3. A manifest that has 2 pointers pointing to the arm64 image and the amd64 image

The manifest will be tagged `quay.io/devfile/oci-registry:next` for example, so when a user runs something like `docker pull quay.io/devfile/oci-registry:next` the manifest is going to allow them to receive the proper image for their architecture as it will redirect the pull to the proper image. I am on arm64 so without needing to pass anything if I run `docker pull quay.io/devfile/oci-registry:next` I will pull an arm64 image.

The arm64 and amd64 images have the tag `amd64` and `arm64` because that also enables someone to pull that particular image if they want as well.



> Do you think it might be better to assign the "$BASE_REPO:$BASE_TAG" to a variable so we can use the variable instead?

Yeah I can implement that change!

oci-registry/build-multi-arch.sh
Jdubrick: This might be a local issue with your Docker, the same thing was happening until I added QEMU for Docker here: https://github.com/devfile/registry-support/pull/240/commits/d8e62391a01b68ba740c54faa831ad87a260cb64

I believe it is an issue with emulation or something along those lines, can you run `podman run --rm --privileged tonistiigi/binfmt` or `docker run --rm --privileged tonistiigi/binfmt` and check what architectures are under the "supported" section like so:
```
"supported": [
    "linux/arm64",
    "linux/amd64",
    "linux/riscv64",
    "linux/ppc64le",
    "linux/s390x",
    "linux/386",
    "linux/mips64le",
    "linux/mips64"
  ]
```

thepetk: Yeap adding QEMU works :) Do you think it makes sense to add this information somewhere as troubleshooting in case more people face the same issue?

oci-registry/build-multi-arch.sh
thepetk: > The manifest will be tagged quay.io/devfile/oci-registry:next for example, so when a user runs something like docker pull quay.io/devfile/oci-registry:next the manifest is going to allow them to receive the proper image for their architecture as it will redirect the pull to the proper image. I am on arm64 so without needing to pass anything if I run docker pull quay.io/devfile/oci-registry:next I will pull an arm64 image.

Nice thanks for the explanation! IMO I like this approach and I think it makes sense.

oci-registry/build-multi-arch.sh
Jdubrick: Awesome I'm glad that fixed the issue! Since the `build-multi-arch.sh` was initially designed for being run as part of our CI I didn't think to add a readme or anything related to it, but it may be a good idea incase someone wants to run locally. Do you think including a section in the `oci-registry/README` or a comment within `oci-registry/build-multi-arch.sh` itself would be best for this situation?

thepetk: Yeah maybe a testing locally section or something like this I think is sufficient.

thepetk: In order to check the support I've followed the command you've shared. I've installed QEMU with this command:

```
sudo apt-get install -y gcc-arm-linux-gnueabihf libc6-dev-armhf-cross qemu-user-static qemu-system-i386
```

thepetk (approved):
/lgtm tested the scripts locally using my personal quay.io account and they were working correctly.

PR #246, opened 2024-06-11
fix #540

Horiodino:

Fixes #https://github.com/devfile/api/issues/540

**PR acceptance criteria**:

- [x] Test Coverage 

Documentation (WIP)

### DIFF
--- a/index/server/pkg/server/endpoint.go
+++ b/index/server/pkg/server/endpoint.go
@@ -544,8 +544,8 @@ func buildIndexAPIResponse(c *gin.Context, indexType string, wantV1Index bool, p
 		util.FilterDevfileDeprecated(&index, *params.Deprecated, wantV1Index)
 	}
 	if index == nil || len(index) == 0 {
-		c.JSON(http.StatusOK, gin.H{
-			"status": "No samples found in the registry",
+		c.JSON(http.StatusNotFound, gin.H{
+			"status": "No stacks or samples found in the registry",
 		})
 		return
 	}
--- a/index/server/pkg/server/endpoint.go
+++ b/index/server/pkg/server/endpoint.go
@@ -543,6 +543,12 @@ func buildIndexAPIResponse(c *gin.Context, indexType string, wantV1Index bool, p
 	if params.Deprecated != nil {
 		util.FilterDevfileDeprecated(&index, *params.Deprecated, wantV1Index)
 	}
+	if index == nil || len(index) == 0 {
+		c.JSON(http.StatusOK, gin.H{
+			"status": "No samples found in the registry",
+		})
+		return
+	}
 
 	if wantV1Index {
 		index = util.ConvertToOldIndexFormat(index)
--- a/index/server/pkg/server/endpoint.go
+++ b/index/server/pkg/server/endpoint.go
@@ -544,7 +544,7 @@ func buildIndexAPIResponse(c *gin.Context, indexType string, wantV1Index bool, p
 		util.FilterDevfileDeprecated(&index, *params.Deprecated, wantV1Index)
 	}
 	if index == nil || len(index) == 0 {
-		c.JSON(http.StatusNotFound, gin.H{
+		c.JSON(http.StatusOK, gin.H{
 			"status": "No stacks or samples found in the registry",
 		})
 		return


Jdubrick:
All the checks pass in our CI - Are you still experiencing issues building locally @Horiodino 

fyi @michael-valdron

Horiodino:
yes

Horiodino:
yep i can build image and test case is running!

Jdubrick:
> yep i can build image and test case is running!

No issues locally anymore? @Horiodino

Horiodino:
nope

index/server/pkg/server/endpoint.go
```diff
@@ -543,6 +543,12 @@ func buildIndexAPIResponse(c *gin.Context, indexType string, wantV1Index bool, p
 	if params.Deprecated != nil {
 		util.FilterDevfileDeprecated(&index, *params.Deprecated, wantV1Index)
 	}
+	if index == nil || len(index) == 0 {
+		c.JSON(http.StatusOK, gin.H{
```
Jdubrick, line 547:
```suggestion
		c.JSON(http.StatusInternalServerError, gin.H{
```

I'm wondering if this should be a [500](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) instead of a 200 as it not having the samples is technically an error and is referenced as an error in the [original issue](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/500) 


https://go.dev/src/net/http/status.go

wdyt @michael-valdron

Jdubrick (changes requested):

index/server/pkg/server/endpoint.go
michael-valdron: @Jdubrick Maybe [404](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/404) would be more appropriate error code here, since no stacks or samples can be found.

index/server/pkg/server/endpoint.go
```diff
@@ -543,6 +543,12 @@ func buildIndexAPIResponse(c *gin.Context, indexType string, wantV1Index bool, p
...
+			"status": "No samples found in the registry",
```
michael-valdron, line 548: Because this response is returned if index has no entries and not just the samples, the status should mention stacks too:
```suggestion
			"status": "No stacks or samples found in the registry",
```

michael-valdron (changes requested):

index/server/pkg/server/endpoint.go
Jdubrick: I was thinking `500` as it would be the servers fault no samples were loaded, no? I am also okay with `404` though

michael-valdron: > I was thinking `500` as it would be the servers fault no samples were loaded, no? I am also okay with `404` though

That is a good point, but in this case the control logic that provides this response does not guarantee that the stacks and samples are not loaded due to error versus just having an empty registry.

Jdubrick: I'm okay with `404` in that case. 

@Horiodino for reference that is `http.StatusNotFound`

johnmcollier: IMO, in a world where we're working towards more dynamic devfile registries, I'd argue that a 200 code is fine here.

index/server/pkg/server/endpoint.go
```diff
@@ -543,6 +543,12 @@ func buildIndexAPIResponse(c *gin.Context, indexType string, wantV1Index bool, p
 	if params.Deprecated != nil {
 		util.FilterDevfileDeprecated(&index, *params.Deprecated, wantV1Index)
 	}
+	if index == nil || len(index) == 0 {
+		c.JSON(http.StatusNotFound, gin.H{
```
johnmcollier, line 547: TBH, I think this should actually be `http.StatusOK`.

@michael-valdron @Jdubrick It might be unusual to have a registry return no stacks, but IMO, I think 200 is more appropriate here than 404 or 500, especially as we work towards a more dynamic registry in https://github.com/devfile/api/issues/1505

johnmcollier (commented):
TBH, I'd

index/server/pkg/server/endpoint.go
Jdubrick: I'm okay with reverting it back to a `200` response. @Horiodino feel free to revert that change

Jdubrick (approved):
Tested locally - Removed `extraDevfileEntries.yaml` from the `tests` directory and created the registry. After hitting `/index/sample` I was able to receive the correct response.

@michael-valdron since you requested changes I'll wait until you take a look as well before merging

michael-valdron (approved):
/lgtm

Tried locally with an empty index, works as expected in headless:

image

@devfile/devfile-services-team I noticed that when the index empty that the registry viewer is throwing errors:
```
TypeError: devfileJsons[devfileRegistryIndex].map is not a function
    at /app/apps/registry-viewer/dist/.next/server/chunks/951.js:5066:110
    at Array.flatMap (<anonymous>)
    at fetchDevfiles (/app/apps/registry-viewer/dist/.next/server/chunks/951.js:5066:33)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
    at async getServerSideProps (/app/apps/registry-viewer/dist/.next/server/pages/index.js:205:25)
    at async Object.renderToHTML (/app/node_modules/next/dist/server/render.js:507:20)
    at async doRender (/app/node_modules/next/dist/server/base-server.js:720:34)
    at async cacheEntry.responseCache.get.incrementalCache.incrementalCache (/app/node_modules/next/dist/server/base-server.js:837:28)
    at async /app/node_modules/next/dist/server/response-cache/index.js:83:36
```

If we are expecting a devfile registry to have a case with an empty index, there should be a follow issue created to address this on the registry viewer.

michael-valdron:
@Horiodino Changes looks good and merging this now, thanks for your contribution towards this!

Horiodino:
> @Horiodino Changes looks good and merging this now, thanks for your contribution towards this!

:raised_hands:  Looking forward to working more , thanks

PR #248, opened 2024-06-21
WIP: Add lastModified field to generated index file

Jdubrick:

This is the first PR aimed at supporting a `lastModified` query parameter for filtering data. The changes made as part of this PR add functionality for reading a `last_modified.json` file that is added to the `registry` directory during build. The goal is to read these values and assign them to their proper stack/sample versions they belong to within the generated `index` file. 

In addition to this functionality I had to do more data manipulation to allow for the proper setting of these values, it is easier to do this in `go` rather than `bash` so it was done here instead of https://github.com/devfile/registry/pull/429.

Additionally, tests were updated to contain the new field so we can ensure it is being added correctly. 

FYI changes still need to be made on the `index/server` side for the actual endpoint, this is just the `index/generator` work for the index file.

fixes https://github.com/devfile/api/issues/1327

**PR acceptance criteria**:

- [ ] Test Coverage 

Documentation (WIP)

Go tests should pass - The new field was added to the expected index file.

### DIFF
--- a/index/generator/library/library.go
+++ b/index/generator/library/library.go
@@ -122,6 +122,10 @@ func GenerateIndexStruct(registryDirPath string, force bool) ([]schema.Schema, e
 		index = append(index, indexFromExtraDevfileEntries...)
 	}
 
+	index, err = SetLastModifiedValue(index, registryDirPath)
+	if err != nil {
+		return index, err
+	}
 	return index, nil
 }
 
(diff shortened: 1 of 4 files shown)


Jdubrick:
I added the gosec exemption as it is following and cleaning the path the same as the following:
https://github.com/devfile/registry-support/blob/main/index/generator/library/library.go#L412
https://github.com/devfile/registry-support/blob/main/index/generator/library/library.go#L545
https://github.com/devfile/registry-support/blob/main/index/generator/library/library.go#L633

Jdubrick:
https://github.com/devfile/registry-support/actions/runs/9617745553/job/26530169277?pr=248 should be resolved after https://github.com/devfile/registry/pull/429 is merged

Jdubrick:
Going to close - I'll open a new PR with the index/server changes. It will be too much of a headache to have them in separate PRs for tests

PR #249, opened 2024-06-24
Feature: Last Modified Filter Parameter

Jdubrick:

This PR adds an additional filter parameter for stacks/samples on the `GET /v2index` endpoint.

Changes made include:
- Addition of `last_modified.json` and `lastModified` field to test data
- Updated unit tests for proper filtering based on the `lastModified` parameter
- Logic for creation of `index.json` to contain `lastModified` field based off of the `last_modified.json` created through registry build
- Update to generator schema to reflect new field
- Update to generator `index_main.json` to ensure last modified field is added correctly

fixes https://github.com/devfile/api/issues/1327

**PR acceptance criteria**:

- [ ] Test Coverage 

Documentation (WIP)

FYI to reviewers: The linked issue has notes for `GET /index` and `GET /v2index`. The script that generates `last_modified.json` that was merged [here](https://github.com/devfile/registry/pull/429) currently creates this file based on the updated Schema used in `v2index` where there is a `Versions` field that contains `[]Version`. For the `/index` we would need to adjust the script to add the `lastModified` field to the entire stack as a whole and not just the individual stack versions.

- Go tests should pass for `index/generator` and `index/server`
- To test locally you can spin up the registry and perform the following:
    1. `GET /v2index?lastModified=2024-04-29` to see more than 1 stack that has this date
    2. `GET /v2index?lastModified=2024-04-29&name=<name>` to filter even further
    3. `GET /v2index/all`, `GET /v2index/stack` `GET/v2index/sample` to observe the new field added in the response as well

### DIFF
--- a/index/generator/library/library.go
+++ b/index/generator/library/library.go
@@ -24,6 +24,7 @@ import (
 	"path"
 	"path/filepath"
 	"strings"
+	"time"
 
 	devfileParser "github.com/devfile/library/v2/pkg/devfile"
 	"github.com/devfile/library/v2/pkg/devfile/parser"
@@ -725,7 +726,7 @@ func SetLastModifiedValue(index []schema.Schema, registryDirPath string) ([]sche
 		return index, err
 	}
 
-	lastModifiedEntriesMap := make(map[string]map[string]string)
+	lastModifiedEntriesMap := make(map[string]map[string]time.Time)
 
 	for idx := range lastModifiedEntries.Stacks {
 		updateLastModifiedMap(lastModifiedEntriesMap, &lastModifiedEntries.Stacks[idx])
@@ -736,28 +737,34 @@ func SetLastModifiedValue(index []schema.Schema, registryDirPath string) ([]sche
 	}
 
 	for i := range index {
+		var mostCurrentLastModifiedDate time.Time
 		for j := range index[i].Versions {
 			schemaItem := index[i] // a stack or sample
-			versions := schemaItem.Versions[j]
-			versionNum := versions.Version
-			updateSchemaLastModified(&schemaItem, j, lastModifiedEntriesMap[schemaItem.Name][versionNum])
+			version := schemaItem.Versions[j]
+			versionNum := version.Version
+			lastModifiedDate := lastModifiedEntriesMap[schemaItem.Name][versionNum]
+			updateSchemaLastModified(&schemaItem, j, lastModifiedDate)
+			if lastModifiedDate.After(mostCurrentLastModifiedDate) {
+				mostCurrentLastModifiedDate = lastModifiedDate
+			}
 		}
+		// lastModified of a stack or sample will be the date any version of it was last changed
+		index[i].LastModified = mostCurrentLastModifiedDate.Format(time.RFC3339)
 	}
 
 	return index, nil
 }
 
-func updateLastModifiedMap(m map[string]map[string]string, entry *schema.LastModifiedEntry) {
+func updateLastModifiedMap(m map[string]map[string]time.Time, entry *schema.LastModifiedEntry) {
 	_, ok := m[entry.Name]
 	if !ok {
-		m[entry.Name] = make(map[string]string)
+		m[entry.Name] = make(map[string]time.Time)
 	}
-
-	m[entry.Name][entry.Version] = entry.LastModified.Format("2006-01-02") //2006-01-02 is used to turn the date into YYYY-MM-DD format
+	m[entry.Name][entry.Version] = entry.LastModified
 }
 
-func updateSchemaLastModified(s *schema.Schema, versionIndx int, lastModifiedDate string) {
-	s.Versions[versionIndx].LastModified = lastModifiedDate
+func updateSchemaLastModified(s *schema.Schema, versionIndx int, lastModifiedDate time.Time) {
+	s.Versions[versionIndx].LastModified = lastModifiedDate.Format(time.RFC3339)
 }
 
 // In checks if the value is in the array
--- a/index/generator/library/library.go
+++ b/index/generator/library/library.go
@@ -122,6 +122,10 @@ func GenerateIndexStruct(registryDirPath string, force bool) ([]schema.Schema, e
 		index = append(index, indexFromExtraDevfileEntries...)
 	}
 
+	index, err = SetLastModifiedValue(index, registryDirPath)
+	if err != nil {
+		return index, err
+	}
 	return index, nil
 }
 
@@ -703,6 +707,52 @@ func validateStackInfo(stackInfo schema.Schema, stackfolderDir string) []error {
 	}
 
 	return errors
+}
+
+// SetLastModifiedValue adds the last modified value to a pre-created index
+// The last modified dates are contained in a file named last_modified.json that is apart of the registry dir
+func SetLastModifiedValue(index []schema.Schema, registryDirPath string) ([]schema.Schema, error) {
+	bytes, err := os.ReadFile(registryDirPath + "/last_modified.json")
+	if err != nil {
+		return index, err
+	}
+
+	var lastModifiedEntries schema.LastModifiedInfo
+	err = json.Unmarshal(bytes, &lastModifiedEntries)
+	if err != nil {
+		return index, err
+	}
+
+	lastModifiedEntriesMap := make(map[string]map[string]string)
+
+	for _, entry := range lastModifiedEntries.Stacks {
+		updateLastModifiedMap(lastModifiedEntriesMap, &entry)
+	}
+
+	for _, entry := range lastModifiedEntries.Samples {
+		updateLastModifiedMap(lastModifiedEntriesMap, &entry)
+	}
+
+	for _, entry := range index {
+		for idx := range entry.Versions {
+			updateSchemaLastModified(&entry, idx, lastModifiedEntriesMap[entry.Name][entry.Versions[idx].Version])
+		}
+	}
+
+	return index, nil
+}
+
+func updateLastModifiedMap(m map[string]map[string]string, entry *schema.LastModifiedEntry) {
+	_, ok := m[entry.Name]
+	if !ok {
+		m[entry.Name] = make(map[string]string)
+	}
+
+	m[entry.Name][entry.Version] = entry.LastModified.Format("2006-01-02") //2006-01-02 is used to turn the date into YYYY-MM-DD format
+}
+
+func updateSchemaLastModified(s *schema.Schema, versionIndx int, lastModifiedDate string) {
+	s.Versions[versionIndx].LastModified = lastModifiedDate
 }
 
 // In checks if the value is in the array
--- a/index/generator/library/library.go
+++ b/index/generator/library/library.go
@@ -712,7 +712,8 @@ func validateStackInfo(stackInfo schema.Schema, stackfolderDir string) []error {
 // SetLastModifiedValue adds the last modified value to a pre-created index
 // The last modified dates are contained in a file named last_modified.json that is apart of the registry dir
 func SetLastModifiedValue(index []schema.Schema, registryDirPath string) ([]schema.Schema, error) {
-	bytes, err := os.ReadFile(registryDirPath + "/last_modified.json")
+	lastModFile := path.Join(registryDirPath, "last_modified.json")
+	bytes, err := os.ReadFile(lastModFile)
 	if err != nil {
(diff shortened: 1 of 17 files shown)


index/generator/library/library.go
```diff
@@ -705,6 +709,54 @@ func validateStackInfo(stackInfo schema.Schema, stackfolderDir string) []error {
...
+	return index, nil
+}
+
+func updateLastModifiedMap(m map[string]map[string]string, entry *schema.LastModifiedEntry) {
+	_, ok := m[entry.Name]
+	if !ok {
+		m[entry.Name] = make(map[string]string)
+	}
+
+	m[entry.Name][entry.Version] = entry.LastModified.Format("2006-01-02") //2006-01-02 is used to turn the date into YYYY-MM-DD format
```
thepetk, line 753: Shouldn't we include the full datetime here? I think having both date and time can help for sorting purposes in the future.

index/generator/schema/schema.go
```diff
@@ -264,4 +266,16 @@ type Version struct {
 	DeploymentScopes map[DeploymentScopeKind]bool `yaml:"deploymentScopes,omitempty" json:"deploymentScopes,omitempty"`
 	Resources        []string                     `yaml:"resources,omitempty" json:"resources,omitempty"`
 	StarterProjects  []string                     `yaml:"starterProjects,omitempty" json:"starterProjects,omitempty"`
+	LastModified     string                       `yaml:"lastModified,omitempty" json:"lastModified,omitempty"`
```
thepetk, line 269: I think we might need to update the `schema.go` [here](https://github.com/devfile/registry-support/blob/ce8b9b9bed12d4dbafa1726e6ea490f2e3cf1ce7/index/generator/schema/schema.go#L25-L117)

Also after building and testing the new `devfile-index-schema` I've noticed that the `lastModified` is only added to versions. I think it should also be included in the [general schema](https://github.com/devfile/registry-support/blob/ce8b9b9bed12d4dbafa1726e6ea490f2e3cf1ce7/index/generator/schema/schema.go#L144-L166). For versioned stacks it should take the value of the `default` stack version. WDYT?

index/generator/library/library.go
```diff
@@ -705,6 +709,54 @@ func validateStackInfo(stackInfo schema.Schema, stackfolderDir string) []error {
...
+		updateLastModifiedMap(lastModifiedEntriesMap, &lastModifiedEntries.Stacks[idx])
+	}
+
+	for idx := range lastModifiedEntries.Samples {
+		updateLastModifiedMap(lastModifiedEntriesMap, &lastModifiedEntries.Samples[idx])
+	}
+
+	for i := range index {
+		for j := range index[i].Versions {
+			updateSchemaLastModified(&index[i], j, lastModifiedEntriesMap[index[i].Name][index[i].Versions[j].Version])
```
thepetk, line 740: I'm wondering here if it's better to assign the params passed in this function to variables in order the code to be more readable. WDYT?

index/server/pkg/server/endpoint.go
```diff
@@ -551,6 +551,7 @@ func buildIndexAPIResponse(c *gin.Context, indexType string, wantV1Index bool, p
 		maxSchemaVersion := params.MaxSchemaVersion
 		minVersion := params.MinVersion
 		maxVersion := params.MaxVersion
+		lastModified := params.LastModified
```
thepetk, line 554: Here I'd expect to be able to filter for min-max similar to the `minVersion`, `maxVersion`

thepetk (changes requested):
I've added some comments.

Tried to test the registry viewer against the new `devfile-index-base` image. I've noticed that some stacks were not visible with the current index:

image
image

index/generator/library/library.go
Jdubrick: I set it to only `YYYY-MM-DD` as per https://github.com/devfile/api/issues/959#issuecomment-1266046556

Jdubrick:
> I've added some comments.
> 
> Tried to test the registry viewer against the new `devfile-index-base` image. I've noticed that some stacks were not visible with the current index. Not sure if there'll be any impact on the current changes and the viewer:
> 
> image

Hmm that's interesting, it shouldn't be affecting the default index that is being returned at all. Are you running it with the test data? If so then it is displaying properly as far as I can tell from the screenshot: https://github.com/Jdubrick/registry-support/tree/main/tests/registry

index/generator/library/library.go
thepetk: Hmm, not sure about that, I think it makes more sense to keep all the information. IMO the server should be serving the full info and it's up to the viewer apply a different format.

index/generator/schema/schema.go
thepetk: I think that's a good point here. The "lastModified" for the stack should be the actual "lastModified" so after a second thought the stack `lastModified` should have the value of the latest modified version. WDYT?

index/generator/library/library.go
Jdubrick: I am okay with the server returning the full info in the response, for the parameter passed to the endpoint however I feel it is best to leave ti in `YYYY-MM-DD` and the server can do the necessary jobs to return the appropriate index. I don't know how much value the endpoint serves if it is looking for precision down to the millisecond to filter, wdyt?

index/generator/schema/schema.go
Jdubrick: If we attach `lastModified` to the entire stack, what are we looking for in terms of the logic? If the stack shares a `lastModified` date with the latest modified version and that date is sent as a parameter it will send the same data back as it is configured now. Is the goal of attaching the `lastModified` to the entire stack as well just to include that information for the viewer since the response in terms of stack versions is going to be the same

Jdubrick: I implemented the change we talked about above, just trying to work through why the test for checking the generated index is failing. Jsondiff is saying they are equal but still failing it

thepetk: > If we attach `lastModified` to the entire stack, what are we looking for in terms of the logic? If the stack shares a `lastModified` date with the latest modified version and that date is sent as a parameter it will send the same data back as it is configured now. Is the goal of attaching the `lastModified` to the entire stack as well just to include that information for the viewer since the response in terms of stack versions is going to be the same

My main point for this is to try to keep consistency in the structure of the `v2index/all` call. That said when someone cass the endpoint they get:

```
stack:
{
    "name": "code-with-quarkus",
    "displayName": "Basic Quarkus",
    "description": "A simple Hello World Java application using Quarkus",
    "type": "sample",
    "tags": [
      "Java",
      "Quarkus"
    ],
    "icon": "https://design.jboss.org/quarkus/logo/final/SVG/quarkus_icon_rgb_default.svg",
    "projectType": "quarkus",
    "language": "java",
    "versions": [
      {
        "version": "1.1.0",
        "schemaVersion": "2.2.0",
        "default": true,
        "git": {
          "remotes": {
            "origin": "https://github.com/devfile-samples/devfile-sample-code-with-quarkus.git"
          }
        },
        "description": "java quarkus with devfile v2.2.0",
        "lastModified": "2024-04-29"
      },
      {
        "version": "1.0.0",
        "schemaVersion": "2.0.0",
        "git": {
          "remotes": {
            "origin": "https://github.com/elsony/devfile-sample-code-with-quarkus.git"
          }
        },
        "description": "java quarkus with devfile v2.0.0",
        "lastModified": "2024-02-15"
      }
    ]
}

sample:
{
  "name": "java-springboot-basic",
  "displayName": "Basic Spring Boot",
  "description": "A simple Hello World Java Spring Boot application using Maven",
  "type": "sample",
  "tags": [
    "Java",
    "Spring"
  ],
  "icon": "https://raw.githubusercontent.com/devfile-samples/devfile-stack-icons/main/spring.svg",
  "projectType": "springboot",
  "language": "java",
  "git": {
    "remotes": {
      "origin": "https://github.com/devfile-samples/devfile-sample-java-springboot-basic.git"
    }
  }
}
```
I'm thinking from a client perspective (having as an example the `registry-viewer`) it might be useful to include this information to all the items that someone is fetching through the call.

For the community registry this might not be a problem as [all of our samples have versions](https://github.com/devfile/registry/blob/main/extraDevfileEntries.yaml) but in case a registry has samples without versions the `lastModified` info won't be available to its viewer.

Jdubrick: A change may need to be done [here](https://github.com/devfile/registry/blob/main/.ci/generate_last_mod_file.sh) to get the `lastModified` of a sample that has no versions set unless we want to have a default value (if no versions).

Jdubrick: @thepetk with the most recent changes you might see that if a sample has no versions then the `lastModified` date attached to it is the default `time.Time` value. If we want to set that to be the last time the sample was modified we will need to edit the script [here ](https://github.com/devfile/registry/blob/main/.ci/generate_last_mod_file.sh) I believe. This PR is quite large I'm not sure if we want to incorporate that here or open a follow up issue?

index/server/pkg/server/endpoint.go
Jdubrick: `minLastModified` and `maxLastModified` are now available. You can access them with `/v2index/all?minLastModified=<date>&maxLastModified=<date>` or any variation of `/v2index` with any combination of `minLastModified` or `maxLastModified`

index/generator/schema/schema.go
thepetk: Yeah +1 and let's take it on a discussion, because I feel there are multiple approaches here

index/server/pkg/server/endpoint.go
thepetk: Nice! Let me test!

index/generator/library/library.go
```diff
@@ -705,6 +710,63 @@ func validateStackInfo(stackInfo schema.Schema, stackfolderDir string) []error {
 	return errors
 }
 
+// SetLastModifiedValue adds the last modified value to a pre-created index
+// The last modified dates are contained in a file named last_modified.json that is apart of the registry dir
+/* #nosec G304 -- lastModFile is produced from filepath.Join which cleans the input path */
+func SetLastModifiedValue(index []schema.Schema, registryDirPath string) ([]schema.Schema, error) {
```
thepetk, line 716: I think a test for this function is a nice addition. WDYT?

index/server/pkg/server/endpoint.go
thepetk: I can confirm the `minLastModified` & `maxLastModified` are working :)

index/generator/schema/schema.go
thepetk: After checking the latest, the non-versioned field works :) One thing we might also need to think about is if we want to serve a specific timezone & format, as we currently serve multiple formats.

For example I see:
* `2023-04-08T11:51:08Z`
* `2024-05-13T12:32:02+02:00`

index/generator/library/library.go
Jdubrick: I can add a test for that, I thought I did but maybe I overlooked it!

It won't let me reply to your comment below but I think `2023-04-08T11:51:08Z` and `2024-05-13T12:32:02+02:00` are in the same format (RFC3339 in Go, which is a strict ISO8601). The `Z` indicates UTC and the `+0200` indicates it is 2 hours offset from UTC. I believe instead of stating `+000` it shorthands it to `Z`.

Jdubrick: I think with the `time` pkg in go `RFC3339` would be the best format for achieving `ISO8601` but it includes the timezone (and when the `last_modified.json` file is created on the registry side it includes the timezone as well from git). I think to stick to what we talked about earlier with the server providing all the information it could be left up to the frontend to format it down without a timezone if necessary? wdyt

thepetk: > I think with the `time` pkg in go `RFC3339` would be the best format for achieving `ISO8601` but it includes the timezone (and when the `last_modified.json` file is created on the registry side it includes the timezone as well from git). I think to stick to what we talked about earlier with the server providing all the information it could be left up to the frontend to format it down without a timezone if necessary? wdyt

Yup for me is fine. It's strange from github perspective that those datetime values have such differences.

Jdubrick:
> /lgtm and after merging this PR I think we should create & prioritize an issue so the non-versioned stacks can also have the lastModified field with accurate data :)
> 
> Very nice feature!

Thanks Fanis! FYI I opened https://github.com/devfile/api/issues/1607 and will work on it once refined

PR #253, opened 2024-08-16
Update index/generator version for registry-library

Jdubrick:

This PR accomplishes the following:

- `index/generator` reference updated to https://github.com/devfile/registry-support/commit/cf509ccd1a6b7a5a2d1c39563c95935175268c25 as it pulls in security fix
- `go mod vendor` ran

Part of https://github.com/devfile/api/issues/1620

**PR acceptance criteria**:

- [ ] Test Coverage 

Documentation (WIP)


thepetk (approved):
/lgtm after all checks are passing.

I'm curious, I see it also updates the indirect dependencies of `k8s.io` to `v0.29.2` (comes after the general bump we did for all modules). I don't expect any breaking changes as they would have been seen in the direct dependencies already.

Jdubrick:
> /lgtm after all checks are passing.
> 
> I'm curious, I see it also updates the indirect dependencies of `k8s.io` to `v0.29.2` (comes after the general bump we did for all modules). I don't expect any breaking changes as they would have been seen in the direct dependencies already. Just thinking in the future we might want to include this in a list of items when we do similar bumps across all modules/projects

Yeah I initially saw that too, I think it should be okay since the most breaking change of that bump was the Go version I believe? And this already uses `1.21`.

+1 on taking note for when we bump other projects though.

PR #285, opened 2025-08-04
Updating Go modules from 1.21 to 1.24

JslYoon:
# Description of Changes
Updated go modules from 1.21/1.22 to 1.24:
    
Update registry-library/go.mod to Go 1.24

Update index/generator/go.mod to Go 1.24

Update index/server/go.mod to Go 1.24

Update tests/integration/go.mod to Go 1.24

Update index/server/Dockerfile to Go 1.24

Update oci-registry/Dockerfile to Go 1.24

Update tests/integration/Dockerfile to Go 1.24

Update .ci/Dockerfile to Go 1.24

Update .github/workflows/ci.yaml to Go 1.24

Update .github/workflows/codecov.yaml to Go 1.24

Update .github/workflows/pushimage-next.yaml to Go 1.24

Do code refactoring where needed

Update Go tag under [README](https://github.com/devfile/registry-support/blob/main/README.md)

# Related Issue(s)
https://github.com/devfile/api/issues/1710 && https://github.com/devfile/api/issues/1700

# Acceptance Criteria

### Tests
- [o] Test Coverage 
    - Are your changes sufficiently tested, and are any applicable test cases added or updated to cover your changes?

PR #291, opened 2025-08-19
Revert "Updating Go modules from 1.21 to 1.24"

JslYoon:
Reverts devfile/registry-support#285
Blocked because of Openshift CI Go  1.24 isn't supported


michael-valdron:
@JslYoon found a workable solution for https://github.com/openshift/release/pull/68063 to move forward with Go 1.24.

